PR #3, opened 2019-09-18
Update Dockerfile to use buster

cpretzer:

This commit updates the Dockerfile to use the buster slim image `debian:buster-20190910-slim`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,9 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
         procps \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    && update-alternatives --set iptables /usr/sbin/iptables-legacy \
+    && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 COPY LICENSE /linkerd/LICENSE
 COPY --from=golang /out/linkerd2-proxy-init /usr/local/bin/proxy-init
 ENTRYPOINT ["/usr/local/bin/proxy-init"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ARG TARGETARCH
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -o /out/linkerd2-proxy-init -mod=readonly -ldflags "-s -w" -v
 
 ## package runtime
-FROM --platform=$TARGETPLATFORM debian:buster-20190910-slim
+FROM --platform=$TARGETPLATFORM debian:buster-20200514-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ARG TARGETARCH
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -o /out/linkerd2-proxy-init -mod=readonly -ldflags "-s -w" -v
 
 ## package runtime
-FROM --platform=$TARGETPLATFORM debian:stretch-20190812-slim
+FROM --platform=$TARGETPLATFORM debian:buster-20190910-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
--- a/integration_test/iptables/iptablestest-lab.yaml
+++ b/integration_test/iptables/iptablestest-lab.yaml
@@ -83,6 +83,12 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: false
           runAsUser: 0
+        volumeMounts:
+        - mountPath: /run
+          name: linkerd-proxy-init-xtables-lock
+      volumes:
+      - emptyDir: {}
+        name: linkerd-proxy-init-xtables-lock
 ---
 apiVersion: v1
 kind: Service
@@ -132,6 +138,12 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: false
           runAsUser: 0
+        volumeMounts:
+        - mountPath: /run
+          name: linkerd-proxy-init-xtables-lock
+      volumes:
+      - emptyDir: {}
+        name: linkerd-proxy-init-xtables-lock
 ---
 apiVersion: v1
 kind: Pod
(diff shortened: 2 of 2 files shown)


samlabs821:
debian buster uses nftables, which makes old rules incompatible

Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY . .
 RUN CGO_ENABLED=0 GOOS=linux go build -o /out/linkerd2-proxy-init -mod=readonly -ldflags "-s -w" -v
 
 ## package runtime
-FROM debian:stretch-20190812-slim
+FROM debian:buster-20190910-slim
```
joakimr-axis, line 8: Should this perhaps be `buster-**20200514**-slim` in order to match the buster version used in the [linkerd2](https://github.com/linkerd/linkerd2) repo (https://github.com/linkerd/linkerd2/pull/4542/files)?

Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY . .
...
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
```
joakimr-axis, line 11: Try adding

```
RUN update-alternatives --set iptables /usr/sbin/iptables-legacy \
    && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
```

here to make current use of `iptables` work with buster.
(See https://github.com/linkerd/linkerd2/pull/4567 for more info.)

cpretzer:
Tested with `buster-20200514-slim` and the init containers started with the expected output.

The dashboard and linkerd command line output (stat, tap, etc.) all look correct.

alpeb:
@cpretzer would you apply the iptables-legacy change suggested [above](https://github.com/linkerd/linkerd2-proxy-init/pull/3#discussion_r436845445)? It's necessary for older k8s version, it seems.

cpretzer:
@alpeb I found the same error testing on multiple kubernetes `1.14`, `1.16`, and `1.18`.

I did some digging and found [an issue](https://github.com/kubernetes/kubernetes/commit/0cfef01a443429edcb95c8e5952edd68f2a32cf8) in the Kubernetes repository which describes similar behavior. In that PR they address the issue by creating a `hostPath` to write `/run/xtables.lock`.

kind also [uses](https://github.com/kubernetes-sigs/kind/pull/977) `hostPath`.

I don't think we want to use `hostPath`, so I tested by adding an `emptyDir` volume to a deployment that I manually injected the proxy and proxy-init into and that enabled the container to start. 

If I understand what `iptables` is doing, it's using `xtables.lock` to block other `iptables` processes from updating the current `iptables` rules until while it completes its work. The problem with using `emptyDir` is that other init containers might expect to use this as well and using an `emptyDir` basically just lets the proxy-init container write to memory so that it can execute its commands, which basically isolates the `xtables.lock` for the pod. As a result, if something else is calling `iptables` this lock is meaningless.

I see that we use `hostPath` volumes for the [cni-plugin](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2-cni/templates/cni-plugin.yaml#L230), but adding a `hostPath` volume for `proxy-init` feels like a pretty significant change to the functionality.

What do you think is the right approach here? Is there another option that I've overlooked?

cpretzer:
Okay, I figured out what is going on: `readOnlyRootFileSystem: true` is set in the `securityContext` of the `proxy-init` container.

The integration tests haven't been updated to match the `_proxy-init.tpl` file, so they don't include `readOnlyRootFileSystem` among other settings.

Here is the output from the template in one of my service pods:

```yaml
   initContainers:
...
        name: linkerd-init
...
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            add:
            - NET_ADMIN
            - NET_RAW
          privileged: false
          readOnlyRootFilesystem: true
          runAsNonRoot: false
          runAsUser: 0
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: FallbackToLogsOnError
...
```

And the iptablestest-lab.yaml has:
```yaml
      initContainers:
      - name: linkerd-init
...
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
          privileged: false
          runAsNonRoot: false
          runAsUser: 0
...
```

Using `readOnlyRootFileSystem: true` is the right thing to do, so we need to find a way to make `/run` writable, and I haven't thought of a better way other than the volumeMounts.

alpeb:
Thanks for the extra research @cpretzer :+1: 

From my understanding, `readOnlyRootFileSystem` refers to the container's ability to write to its top layer, which is still separate from the host's file system (I might be wrong here, my understanding of containers is limited). In which case if we set that to true, the pod won't fail but the locking mechanism will be moot, just like with the `emptyDir` option you mentioned.
Although it turns out `readOnlyRootFileSystem: true` is the solution [others](https://github.com/istio/istio/blob/5cf554543ed9c1f28012ff621081b8e1a9ea3b1b/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml#L72-L82) are relying on.

It also seems to me the only way to do this while keeping proper locking is to rely on `hostPath`. We'd have to update our [PSP](https://github.com/linkerd/linkerd2/blob/82e91382b706294853e4c69de3573cf93beb47bf/charts/linkerd2/templates/psp.yaml#L53-L59) to allow that. We can use the PSP `allowedHostPath` rule to limit the path that is writable, but I'm seeing a [scary warning in the docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems) saying that's easily abused.

I'm realizing we're not honoring that lock file in CNI given we're relying on the default `readOnlyRootFileSystem: false` [there](https://github.com/linkerd/linkerd2/blob/82e91382b706294853e4c69de3573cf93beb47bf/charts/linkerd2-cni/templates/cni-plugin.yaml#L157-L237). Perhaps the easiest is to do the same here?

CC @grampelberg

alpeb:
As to why we're observing this behavior now even though we're still using iptables[-legacy], it's because we're also upgrading from iptables `v1.6.0` to `v1.8.2`, and in `v1.6.2` a change was introduced forcing the usage of that lock.

From this [change](https://git.netfilter.org/iptables/commit/?id=80d8bfaac9e2430d710084a10ec78e68bd61e6ec):
> Existing systems that depended on things working mostly correctly
even if there was no lock might be affected by this change

alpeb:
@cpretzer This is running fine for me when I run the control plane in linkerd/linkerd2#4692 using the proxy-init image produced here. The integration test is hanging presumably because the volume changes from linkerd/linkerd2#4692 need to be applied here as well to the init-containers in the pods in `iptablestest-lab.yaml` in order to avoid the same `/run/xtables.lock` issue.

cpretzer:
thanks @alpeb working on this now

PR #10, opened 2020-05-14
Add close wait timeout flag

adleong:
We add a `timeout-close-wait-secs` flag to the proxy-init process that sets the `net.netfilter.nf_conntrack_tcp_timeout_close_wait`.  Note that this will fail if the container is not `privileged`.

Signed-off-by: Alex Leong <alex@buoyant.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ FROM debian:stretch-20190812-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
+        procps \
     && rm -rf /var/lib/apt/lists/*
 COPY LICENSE /linkerd/LICENSE
 COPY --from=golang /out/linkerd2-proxy-init /usr/local/bin/proxy-init
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"log"
 	"os/exec"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


cmd/root.go
```diff
@@ -46,6 +49,18 @@ func NewRootCmd() *cobra.Command {
 		Short: "proxy-init adds a Kubernetes pod to the Linkerd service mesh",
 		Long:  "proxy-init adds a Kubernetes pod to the Linkerd service mesh.",
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			if options.TimeoutCloseWaitSecs != 0 {
```
kleimkuhler, line 53: Should we assert the value here is greater than 0? Or at least: `if options.TimeoutCloseWaitSecs > 0`

Dockerfile
```diff
@@ -9,6 +9,7 @@ FROM debian:stretch-20190812-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
+        procps \
```
kleimkuhler, line 12: Just curious: can we not set the value without this?

cmd/root.go
adleong: It wouldn't hurt to be super defensive like that, but hopefully not necessary.  The injector code which sets this flag sets it from a go `time.Duration` which cannot be negative.

Dockerfile
adleong: this gives us the `sysctl` command

cmd/root.go
```diff
@@ -46,6 +49,18 @@ func NewRootCmd() *cobra.Command {
...
+				sysctl := exec.Command("sysctl", "-w",
+					fmt.Sprintf("net.netfilter.nf_conntrack_tcp_timeout_close_wait=%d", options.TimeoutCloseWaitSecs),
+				)
+				out, err := sysctl.CombinedOutput()
+				fmt.Println(string(out))
```
alpeb, line 58: Can you use `log.Println` like in `iptables.go`? So that in both places we output to stderr.

PR #11, opened 2020-06-04
Add GitHub Actions for CI Process

aliariff:
This workflow already tested in my local fork.

https://github.com/aliariff/linkerd2-proxy-init/pull/1/checks

Signed-off-by: Ali Ariff <ali.ariff12@gmail.com>

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,0 +1,32 @@
+name: Integration tests
+on:
+  pull_request: {}
+  push:
+    paths-ignore:
+    - '*.md'
+    - '**/*.md'
+    branches:
+    - master
+jobs:
+  kind_integration_tests:
+    name: KinD integration tests
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Setup KinD
+      uses: engineerd/setup-kind@v0.4.0
+      with:
+        version: v0.8.1
+    - name: Docker build
+      run: |
+        docker build -t gcr.io/linkerd-io/proxy-init:latest .
+        docker build -t buoyantio/iptables-tester:v1 -f integration_test/iptables/Dockerfile-tester ./integration_test
+    - name: Load image into the local KinD cluster
+      run: |
+        kind load docker-image gcr.io/linkerd-io/proxy-init:latest
+        kind load docker-image buoyantio/iptables-tester:v1
+    - name: Run integration tests
+      run: |
+        cd integration_test
+        ./run_tests.sh
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -31,4 +31,4 @@ jobs:
     - name: Run integration tests
       run: |
         cd integration_test
-        ./run_tests.sh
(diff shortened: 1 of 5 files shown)


aliariff:
@cpretzer Please take a look

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,32 @@
...
+on:
+  pull_request: {}
+  push:
+    paths-ignore:
+    - '*.md'
+    - '**/*.md'
+    branches:
+    - master
+jobs:
+  kind_integration_tests:
```
cpretzer, line 11: Let's chat about using KInD. I attempted to install KInD on an arm server and it required building k8s, as well.

I recall you saying that the Docker build portion would be fine on arm systems because the Dockerfile will detect the architecture. If `engineerd/setup-kind` will also detect the architecture, then we should be safe to use this

.github/workflows/static_checks.yml
```diff
@@ -0,0 +1,34 @@
+name: Static checks
+on:
+  pull_request: {}
+  push:
+    paths-ignore:
+    - '*.md'
+    - '**/*.md'
+    branches:
+    - master
+jobs:
```
cpretzer, line 10: What do you think about adding a `go_dependencies` job similar to the `linkerd/linkerd2` [repo](https://github.com/linkerd/linkerd2/blob/master/.github/workflows/static_checks.yml#L11)?

We'd have to add a script as well, but I think that's okay. The job and the script can be addressed in another PR.

iptables/iptables.go
```diff
@@ -330,7 +330,7 @@ func makeShowAllRules() *exec.Cmd {
 func asDestination(portRange ports.PortRange) string {
 	if portRange.LowerBound == portRange.UpperBound {
 		return fmt.Sprintf("%d", portRange.LowerBound)
-	} else {
-		return fmt.Sprintf("%d:%d", portRange.LowerBound, portRange.UpperBound)
 	}
+
+	return fmt.Sprintf("%d:%d", portRange.LowerBound, portRange.UpperBound)
```
cpretzer, line 335: Nice efficiency 😄

cpretzer:
GitHub actions don't support arm architecture, so we need to find a solution to run the arm images for integration tests.

@ihcsim I know you've done some work with CI and github actions. Do you have a recommendation here? There's an option to use a self-hosted runner, but it comes with a warning to not use self-hosted runners for public repositories.

The problem we want to solve is to run the integration tests on arm 64 architecture in addition to other architectures

.github/workflows/static_checks.yml
aliariff: I don't think this is necessary, because in `linkerd/linkerd2` repo the `go_dependencies` job is making sure the Dockerfile is using the correct builder image.
In this repo, there is no such case to make sure the builder image.

aliariff:
@cpretzer @ihcsim 

I found terraform configuration file from official packet

https://github.com/packet-labs/packet-multiarch-k8s-terraform (using k8s)
https://github.com/packet-labs/packet-k3s (using k3s)

We can use this similarly like what we already have in https://github.com/linkerd/linkerd2-action-gcloud. Spin up a new cluster on demand, and destroy it after finished.

So we can perform an integration test on the ARM64 machine when we need it.

PR #12, opened 2020-06-09
Add release workflow

aliariff:
@cpretzer 

Tested manually in my local fork:
https://github.com/aliariff/linkerd2-proxy-init/pull/4/checks?check_run_id=800916941
Using the docker hub registry and my user & pass.
The resulting images are in https://hub.docker.com/r/etni35/proxy-init/tags
image

This version only can be tested properly when we create a new release tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: Release
+on:
+  push:
+    tags:
+    - "*"
+jobs:
+  buildx:
+    name: Build & Push Multi Arch Images
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Configure gcloud
+      uses: linkerd/linkerd2-action-gcloud@v1.0.2
+      with:
+        cloud_sdk_service_account_key: ${{ secrets.CLOUD_SDK_SERVICE_ACCOUNT_KEY }}
+        gcp_project: ${{ secrets.GCP_PROJECT }}
+        gcp_zone: ${{ secrets.GCP_ZONE }}
+    - name: Set up Docker Buildx
+      uses: crazy-max/ghaction-docker-buildx@v3
+    - name: Docker Buildx (build)
+      run: make images
+    - name: Docker Buildx (push)
+      run: make push
+    - name: Docker Check Manifest
+      run: make inspect-manifest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - name: Checkout code
-      # actions/checkout@v2
+      # actions/checkout@v2.0.0
       uses: actions/checkout@722adc6
     - name: Configure gcloud
       # linkerd/linkerd2-action-gcloud@v1.0.1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,15 +9,18 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      # actions/checkout@v2
+      uses: actions/checkout@722adc6
     - name: Configure gcloud
-      uses: linkerd/linkerd2-action-gcloud@v1.0.2
+      # linkerd/linkerd2-action-gcloud@v1.0.1
+      uses: linkerd/linkerd2-action-gcloud@308c4df
       with:
         cloud_sdk_service_account_key: ${{ secrets.CLOUD_SDK_SERVICE_ACCOUNT_KEY }}
         gcp_project: ${{ secrets.GCP_PROJECT }}
         gcp_zone: ${{ secrets.GCP_ZONE }}
     - name: Set up Docker Buildx
-      uses: crazy-max/ghaction-docker-buildx@v3
+      # crazy-max/ghaction-docker-buildx@v3.1.0
+      uses: crazy-max/ghaction-docker-buildx@373dafb
     - name: Docker Buildx (build)
       run: make images
     - name: Docker Buildx (push)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,19 @@
 ## compile proxy-init utility
-FROM golang:1.12.9 as golang
+FROM --platform=$BUILDPLATFORM golang:1.12.9 as golang
 WORKDIR /build
+
+# cache dependencies
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+# build
 COPY . .
-RUN CGO_ENABLED=0 GOOS=linux go build -o /out/linkerd2-proxy-init -mod=readonly -ldflags "-s -w" -v
+ARG TARGETARCH
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -o /out/linkerd2-proxy-init -mod=readonly -ldflags "-s -w" -v
 
 ## package runtime
-FROM debian:stretch-20190812-slim
+FROM --platform=$TARGETPLATFORM debian:stretch-20190812-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -49,3 +51,26 @@ tester-image: ## Build docker image for the tester component
...
+images: ## Build multi arch docker images for the project
+	docker buildx build \
+		--platform $(SUPPORTED_ARCHS) \
+		--output "type=image,push=false" \
+		--tag $(REPO):$(VERSION) \
+		--tag $(REPO):latest \
+		.
+
+.PHONY: push
+push: images ## Push multi arch docker images to the registry
```
ihcsim, line 65: This will cause CI to run `docker buildx build` twice, using different `--output` configuration. I think on CI, we should be able to just use `make push`.

Makefile
```diff
@@ -49,3 +51,26 @@ tester-image: ## Build docker image for the tester component
 kind-load: image tester-image ## Load the required image to KinD cluster
 	kind load docker-image $(REPO):latest
 	kind load docker-image $(TESTER_REPO):v1
+
+.PHONY: images
+images: ## Build multi arch docker images for the project
+	docker buildx build \
+		--platform $(SUPPORTED_ARCHS) \
+		--output "type=image,push=false" \
```
ihcsim, line 59: AIUI, since we are building multi-platform images (using builders with `docker-container` driver), this doesn't output any images, right? (See ee9b892c82f6.) So is the purpose of this rule simply to validate the image build? Do we need this target? Or shall we just provide a single target that can build and push the multi-platform images, and we **don't** default `REGISTRY` to `gcr.io/linkerd-io`?

Dockerfile
```diff
@@ -1,11 +1,19 @@
 ## compile proxy-init utility
-FROM golang:1.12.9 as golang
+FROM --platform=$BUILDPLATFORM golang:1.12.9 as golang
```
ihcsim, line 2: Is `$BUILDPLATFORM` necessary?

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+    tags:
+    - "*"
+jobs:
+  buildx:
+    name: Build & Push Multi Arch Images
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout code
+      # actions/checkout@v2
+      uses: actions/checkout@722adc6
```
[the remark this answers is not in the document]
aliariff: This is to guarantee that the code will not change, as @alpeb recommends. And the SHA refer to https://github.com/actions/checkout/releases/tag/v2.0.0

Makefile
```diff
@@ -49,3 +51,26 @@ tester-image: ## Build docker image for the tester component
...
+		--tag $(REPO):$(VERSION) \
+		--tag $(REPO):latest \
+		.
+
+.PHONY: push
+push: images ## Push multi arch docker images to the registry
+	docker buildx build \
+		--platform $(SUPPORTED_ARCHS) \
+		--output "type=image,push=true" \
```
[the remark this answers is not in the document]
aliariff: Possible, but this one just to make a clear distinction between build without pushing and build with pushing.

Makefile
aliariff: Yes, but the second run will be using the cache Overall the whole release workflow finishes in 1 minute.

Dockerfile
aliariff: Yes, because `By default, the target platform of the build request is used.`
Ref: https://docs.docker.com/engine/reference/builder/#from

Makefile
aliariff: yes, this is not output any image, meaning the images will not appear when we call `docker images` but the images will exist in the cache layer.

Yes, validation is the purpose and also clarity between build & push (on CI stages). The push part will just use the previous cache from the build.

aliariff: For the default registry, for now, we use `gcr.io/linkerd-io` but it can be easily changed using env variable
`DOCKER_REGISTRY=<registry> make ....`

Makefile
```diff
@@ -49,3 +51,26 @@ tester-image: ## Build docker image for the tester component
...
 	kind load docker-image $(REPO):latest
 	kind load docker-image $(TESTER_REPO):v1
+
+.PHONY: images
+images: ## Build multi arch docker images for the project
+	docker buildx build \
+		--platform $(SUPPORTED_ARCHS) \
+		--output "type=image,push=false" \
+		--tag $(REPO):$(VERSION) \
+		--tag $(REPO):latest \
```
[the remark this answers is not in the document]
aliariff: I agree, instructions in the readme or directly in `make help` about how to build multi-arch images will be helpful.

.github/workflows/release.yml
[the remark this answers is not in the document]
ihcsim: Can we update the comment to `v2.0.0`? Thanks.

Makefile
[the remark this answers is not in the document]
ihcsim: Maybe we can script it as a different make target (say, `make builder`), and mention in the README that if one encounters the above error message, run `make builder`. What do you think of something like:

```
.PHONY: builder
builder:
   docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
   docker buildx create --name=multiarch-builder --driver=docker-container --platform="${SUPPORTED_ARCHS}" --use
multiarch
   docker buildx inspect multiarch-builder --bootstrap
```

Makefile
ihcsim: Thanks for the clarification. Is it important to have 2 distinct phases in CI, since the build phase is essentially the same as the push, minus the actual push task? 

Just wondering if can we do something like:

```
images:
	docker buildx build \
		--platform $(SUPPORTED_ARCHS) \
		--output "type=image,push=${PUSH_IMAGE}" \
# ...
```
setting `PUSH_IMAGE` to `true` in CI, but `false` by default, and remove the `push` target?

aliariff: important maybe not really, but just for distinction. 

that is a good suggestion, I will change it, to make it DRY.
Regarding removing the `push`, we can still use it, so it's just calling `PUSH_IMAGE=true  make images`. So the caller no needs to know which env variable to set.

Makefile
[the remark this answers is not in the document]
aliariff: That is also possible, I will try it out.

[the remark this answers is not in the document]
aliariff: @ihcsim 
is it necessary to run `docker run --rm --privileged multiarch/qemu-user-static --reset -p yes`?
because in this case, we don't use QEMU, we just utilize golang internal cross-compile features.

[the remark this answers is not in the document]
ihcsim: AIUI, if the Dockerfile has the `RUN` commands, we will need the qemu in order for those commands to be executed during the multi-arch build.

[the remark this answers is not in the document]
ihcsim: AFAICT, the buildx github action has to do the same thing. See https://github.com/crazy-max/ghaction-docker-buildx/blob/44fd46c7833e095f13bb1d8ace66bd821198bfef/src/main.ts#L25.

README.md
```diff
@@ -19,3 +19,11 @@ Then run the tests with:
...
 make integration-test
 ```
+
+# Build Multi-Architecture Docker Images with Buildx
+
+Please refer to https://docs.docker.com/buildx/working-with-buildx/ to enable Buildx.
+
+Run `make builder` to create Buildx instance before starting to build the images.
+
+Run `make images` to start build the images.
```
ihcsim, line 29: Thanks for adding this section. Can we provide more context around why and when a user will need to run `make builder`? How about something like:

In some local environments like Ubuntu, where the default Buildx builder uses the `docker` driver, the `make images` command might fail with the following error:

```
$ make images
multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
Makefile:57: recipe for target 'images' failed
make: *** [images] Error 1
```

To fix this, you can create a new Buildx builder instance by running `make builder`. This command will create a builder that uses the `docker-container` driver that can build multi-platform images. For more information, see the Buildx builder [documentation](https://docs.docker.com/buildx/working-with-buildx/#work-with-builder-instances).

PR #15, opened 2020-06-12
Add makefile

aliariff:


cpretzer:
This should be merged before #12 per @aliariff

cpretzer:
@alpeb Can you have a look at this? 

I chatted with @aliariff about the change and the reasoning behind opting for a Makefile. It's a great way to conform to how builds are done in the other repositories

aliariff:
@alpeb I add some help text. Is that ok?

image

Makefile
```diff
@@ -1,42 +1,50 @@
 DOCKER_REGISTRY ?= gcr.io/linkerd-io
 REPO = $(DOCKER_REGISTRY)/proxy-init
 
+.DEFAULT_GOAL := help
+
+.PHONY: help
+help: ## Show this help
+	@echo 'Info: For integration test using KinD, run: make kind-load && make integration-test'
```
alpeb, line 9: Awesome! I believe `make kind-load integration-test` is enough, no?

Makefile
```diff
@@ -1,42 +1,50 @@
...
+	@echo 'Info: For integration test using minikube, run: make integration-test'
```
alpeb, line 10: I think you can just say something like "For other environments, run `make integration-test` after having uploaded the images" given this might be run in other places like cloud environments.

Makefile
```diff
@@ -0,0 +1,51 @@
...
+REPO = $(DOCKER_REGISTRY)/proxy-init
+TESTER_REPO = buoyantio/iptables-tester
+
+.DEFAULT_GOAL := help
+
+.PHONY: help
+help: ## Show this help
+	@echo 'Info: For integration test using KinD, run: make kind-load && make integration-test'
+	@echo 'Info: For integration test using minikube, run: make integration-test'
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
alpeb, line 11: Loved this technique, I'm gonna steal it :slightly_smiling_face:

Makefile
aliariff: ah yes, you are right

Makefile
aliariff: 😄 
I got it from here https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html

cpretzer:
@aliariff would you mind rebasing from master so that the lint checks pass? We should be able to merge after that

PR #16, opened 2020-06-19
Update the linkerd-init securityContext values

cpretzer:
## What
The `proxy-init` integration tests are returning false positives.

## Why
The `securityContext` configuration integration tests for the proxy-init image has drifted from what is actually generated by `_proxy-init.tpl` in the [linkerd2 repo](https://github.com/linkerd/linkerd2/blob/master/charts/partials/templates/_proxy-init.tpl#L26).

As a result, the tests pass when they should fail after the changes to #3. 

## How
Update `iptablestest-lab.yaml` to include all the values under the `securityContext` key in the chart template.

### DIFF
--- a/integration_test/iptables/iptablestest-lab.yaml
+++ b/integration_test/iptables/iptablestest-lab.yaml
@@ -74,10 +74,13 @@ spec:
         imagePullPolicy: Never
         args: ["-p", "8080",  "-o", "8080", "-u", "2102"]
         securityContext:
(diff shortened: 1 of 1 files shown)


ihcsim:
@cpretzer LGTM. Are the fixes for the linter errors in #3?

cpretzer:
Thanks @ihcsim I committed the lint fixes separately in #17

PR #18, opened 2020-07-08
Add arm64 integration test

aliariff:
@cpretzer 

Workflow tested in my local fork: 
- https://github.com/aliariff/linkerd2-proxy-init/runs/849272469?check_suite_focus=true

It will only run when gets merged. Not available in a pull request.

Required to set the secrets for the ARM64 machine.

This PR is not addressing conflicting images when 2 or more workflow runs at the same time. Followup PR will be created to solve it.

### DIFF
--- a/.github/workflows/arm64_integration_tests.yml
+++ b/.github/workflows/arm64_integration_tests.yml
@@ -1,0 +1,56 @@
+name: ARM64 integration tests
+on:
+  push:
+    paths-ignore:
+    - '*.md'
+    - '**/*.md'
+    branches:
+    - master
+jobs:
+  arm64_integration_tests:
+    name: ARM64 integration tests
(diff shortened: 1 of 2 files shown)


cpretzer:
@aliariff thanks for your work on this. For now, we're going to prioritize the build portion of the GSoC project, and work on the integration tests as a second step.

cpretzer:
@aliariff is it safe to enable these tests now?

aliariff:
Yes, I think we can enable this. But first, we need to provide the ARM machine and set up the proper GitHub secret for it.
We can also just skip these, as we already do the whole integration test on `linkerd2` repo and rely on that one.

aliariff:
Not going to pursue this for now, as its already tested in `linkerd2` repo

PR #19, opened 2020-07-14
temporarily disable push and inspect in release workflow

cpretzer:
## What
As part of the GSoC work, a release workflow was added that requires connectivity to gcr.io using credentials in order to push the images that are built. For now, we are focusing on the build portion of the project and not the release. This means that we'll need to manually push the images until this is re-enabled.

## Why
When a new tag is created, the release workflow runs and fails when it attempts to connect to gcr.io. This is currently blocking creation of new tags in this repo, so this PR disables that process and we'll revisit it later.

## How
The authorization, push, and inspect steps of the release.yaml workflow are commented out in this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,19 +11,19 @@ jobs:
     - name: Checkout code
       # actions/checkout@v2.0.0
       uses: actions/checkout@722adc6
-    - name: Configure gcloud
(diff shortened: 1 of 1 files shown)


cpretzer:
@kleimkuhler yep, that's right. The goal is to be able to build the images. We'll revisit the release part as a smaller bit of separate work.

PR #26, opened 2020-11-10
Fix CNI for nodes using BusyBox binaries (like k3s)

alpeb:
While trying to make the CNI integration tests run on k3s I realized
`nsenter` had an issue parsing arguments.
[Other](https://github.com/rancher/k3s/issues/1434#issuecomment-629315909) projects have stumbled upon this.

The fix is just to separate `nsenter`-specific args from the rest with `--`.


cpretzer (approved):
:+1:

olix0r (approved):

PR #30, opened 2021-01-20
make: Support publishing custom images to private registries

barkardk:
To make it easier to build custom images and deploy from private registries  its nice to be able to push via makefile without having to use buildx. 
Remove the hardcoded latest tagging from the docker image to a configurable tag with value:latest set to default instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 DOCKER_REGISTRY ?= ghcr.io/linkerd
 REPO = $(DOCKER_REGISTRY)/proxy-init
 TESTER_REPO = buoyantio/iptables-tester
+TAG ?= latest
 VERSION ?= $(shell git describe --exact-match --tags 2> /dev/null || git rev-parse --short HEAD)
 SUPPORTED_ARCHS = linux/amd64,linux/arm64,linux/arm/v7
 PUSH_IMAGE ?= false
-
 .DEFAULT_GOAL := help
 
 .PHONY: help
@@ -42,7 +42,13 @@ integration-test: image ## Perform integration test
 ###############
 .PHONY: image
 image: ## Build docker image for the project
-	DOCKER_BUILDKIT=1 docker build -t $(REPO):latest .
+	DOCKER_BUILDKIT=1 docker build -t $(REPO):$(TAG) .
+
+.PHONY: docker-push
+docker-push: ## push to any docker registry
+	make image
+	docker push $(REPO):$(TAG)
+
 
 .PHONY: tester-image
 tester-image: ## Build docker image for the tester component
--- a/README.md
+++ b/README.md
@@ -13,11 +13,27 @@ Start by building and tagging the `proxy-init` image required for the test:
 eval $(minikube docker-env)
 make image
 ```
-
 Then run the tests with:
 
 ```bash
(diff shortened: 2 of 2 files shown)


adleong (approved):

olix0r:
How does differ from the `VERSION` variable and `make push`?

https://github.com/linkerd/linkerd2-proxy-init/blob/51a35572729e3eb2bfff96ec20aaaf3a857deac3/Makefile#L60-L70

barkardk:
I wanted to make use of docker build and docker push without using docker buildx. 
I assumed docker buildx was there for pipeline usages and did not want to ask for its removal.  
 I get the error 
```multiple platforms feature is currently not supported for docker driver. ``` 
when I try to use the `make images`  or `make push` command. 
I just thought is easier to stick to the   
```make image```   as is noted in the README with a custom tag and then just add an entry for a simple docker push

**Comment by olix0r** [comment]
@barkardk makes sense. Do we need both `TAG` and `VERSION`?

**Comment by barkardk** [comment]
@olix0r not sure originally `make image` had latest as hardcoded but VERSION uses git revision. 
```diff
image: ## Build docker image for the project
	DOCKER_BUILDKIT=1 docker build -t $(REPO):latest .
```
I set a new param TAG instead of using VERSION to preserve backwards compatibility.  
Using VERSION makes perfect sense to me if no one has an issue with replacing the default `latest` tag

barkardk:
@olix0r I took out the TAG parameter and replaced it with version.  
 Also added some documentation and combined the make commands so  that docker-push will also build the container.  To avoid the extra step

Makefile
```diff
@@ -42,7 +42,13 @@ integration-test: image ## Perform integration test
 ###############
 .PHONY: image
 image: ## Build docker image for the project
-	DOCKER_BUILDKIT=1 docker build -t $(REPO):latest .
+	DOCKER_BUILDKIT=1 docker build -t $(REPO):$(TAG) .
+
+.PHONY: docker-push
+docker-push: ## push to any docker registry
+	make image
+	docker push $(REPO):$(TAG)
```
olix0r, line 50: I'm a little concerned about recursively invoking `make`... I know this can be fragile.

Can we accomplish the same thing with:
```suggestion
# push to any docker registry
.PHONY: docker-push
docker-push: image
	docker push $(REPO):$(TAG)
```

Makefile
```diff
@@ -42,7 +42,13 @@ integration-test: image ## Perform integration test
 ###############
 .PHONY: image
 image: ## Build docker image for the project
-	DOCKER_BUILDKIT=1 docker build -t $(REPO):latest .
+	DOCKER_BUILDKIT=1 docker build -t $(REPO):$(TAG) .
```
olix0r, line 45: perhaps the makefile should declare `DOCKER_BUILDKIT = 1` globally above?

olix0r (commented):
Thanks for humoring me ;)

kleimkuhler:
Hey @barkardk just checking in on this. Do you still have the interest/time to address the remaining comments on this and get it merged?

alpeb:
Closing this one for lack of activity. @barkardk you're welcomed to reopen if you're still interested.

PR #31, opened 2021-01-21
changes to run as non root, need to go hand in had with helm changes …

pankajmt:
…to make runAsNonRoot and runAsUser configurable

Signed-off-by: Pankaj Tolani <pankaj.tolani@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,16 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         iptables \
         procps \
+        libcap2-bin \
     && rm -rf /var/lib/apt/lists/* \
     && update-alternatives --set iptables /usr/sbin/iptables-legacy \
     && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 COPY LICENSE /linkerd/LICENSE
 COPY --from=golang /out/linkerd2-proxy-init /usr/local/bin/proxy-init
+
+RUN setcap cap_net_raw,cap_net_admin+eip /usr/sbin/xtables-legacy-multi
+RUN touch /run/xtables.lock && chmod 0666 /run/xtables.lock
+RUN groupadd -r linkerd && useradd --uid 76543 -r -g linkerd linkerd
+USER linkerd
+
 ENTRYPOINT ["/usr/local/bin/proxy-init"]
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -351,7 +351,7 @@ func makeRedirectChainForOutgoingTraffic(chainName string, redirectChainName str
 }
 
 func makeShowAllRules() *exec.Cmd {
-	return exec.Command("iptables-save")
+	return exec.Command("iptables", "-t", "nat", "-L")
 }
 
 // asDestination formats the provided `PortRange` for output in commands.


cpretzer (approved):
This is working for me when I test locally.

Would be good to get this into an edge this week. WDYT @adleong

iptables/iptables.go
```diff
@@ -351,7 +351,7 @@ func makeRedirectChainForOutgoingTraffic(chainName string, redirectChainName str
 }
 
 func makeShowAllRules() *exec.Cmd {
-	return exec.Command("iptables-save")
+	return exec.Command("iptables", "-t", "nat", "-L")
```
alpeb, line 354: Can you please explain the reasoning behind this change?

alpeb: Never mind, I've just say the output is much clearer this way :+1:

alpeb:
Closes linkerd/linkerd2#5505

alpeb (commented):
If I set the init container security context with `runAsNonRoot: true`, it then errors with:
```
Error: container has runAsNonRoot and image has non-numeric user (linkerd), cannot verify user is non-root
```
Is it really necessary to define USER here? If so, can we instead try using its UID directly?

iptables/iptables.go
pankajmt: running that command looks like was hard to do without root. and this command still meets the intended objective.

pankajmt:
> If I set the init container security context with `runAsNonRoot: true`, it then errors with:
> 
> ```
> Error: container has runAsNonRoot and image has non-numeric user (linkerd), cannot verify user is non-root
> ```
> 
> Is it really necessary to define USER here? If so, can we instead try using its UID directly?

runAsNonRoot: true
runAsUser: 76543

works

Dockerfile
```diff
@@ -18,9 +18,16 @@ RUN apt-get update \
...
+        libcap2-bin \
     && rm -rf /var/lib/apt/lists/* \
     && update-alternatives --set iptables /usr/sbin/iptables-legacy \
     && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 COPY LICENSE /linkerd/LICENSE
 COPY --from=golang /out/linkerd2-proxy-init /usr/local/bin/proxy-init
+
+RUN setcap cap_net_raw,cap_net_admin+eip /usr/sbin/xtables-legacy-multi
+RUN touch /run/xtables.lock && chmod 0666 /run/xtables.lock
+RUN groupadd -r linkerd && useradd --uid 76543 -r -g linkerd linkerd
```
adleong, line 30: If I'm understanding this correctly, the UID is hardcoded here to 76543 but is configurable in the associated k8s manifest.  Doesn't this mean that configuring the uid to anything other than 76543 would cause this to break?  Ideally, if this is to be configurable, it should be configured in both places.

pankajmt: yeah, given it will be hard to make it configurable here, make it hard coded there?

adleong: I think so

adleong:
Closing due to inactivity.

pankajmt:
Fair enough. I have moved onto something else and struggling to find time to close this off. Hopefully soon I will come back to finish this. But it would be a good value add to linkerd itself to be able to run more secure. Istios and others dont.

PR #34, opened 2021-03-08
Log to stdout instead of stderr

mo4islona:
All  unstructured logs collected from stderr marked usually with error level in tools like fluentd or similar.
Every pod restart causes error spam in monitoring tools though it is a initialisation debug logs - and it is frustrating.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"log"
 	"os"
 
 	"github.com/linkerd/linkerd2-proxy-init/cmd"
 )
 
 func main() {
+	log.SetOutput(os.Stdout)
+
 	if err := cmd.NewRootCmd().Execute(); err != nil {
 		os.Exit(1)
 	}


kleimkuhler (commented):
@mo4islona I'd like to confirm there should not be any change in logs after a default install, correct? Is this just picked up in tools like fluentd? Any chance you could provide an example with the change in output that you see?

For example, after building this locally and installing, I don't see any difference in the logs output (`kubectl logs -n linkerd linkerd-controller-... linkerd-init`).

mo4islona:
There is no any changes in logs made this PR. It just modify stream which is used to dumps logs.
K8s reads both stdout and stderr streams, so you will not see any changes from k8s side.

But if you using linkerd in Google Could Platform there is a problem.
Take a look on these lines in official k8s addon for GCP
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-gcp/fluentd-gcp-configmap.yaml#L104

```
severity ${record['severity'] || if record['stream'] == 'stderr' then 'ERROR' else 'INFO' end}
```

These proxy init logs are really unstructured. Fluentd cant parse log record, so it looks to the stream in which logs are dumped. In case of stdout it set INFO level, otherwise ERROR

The root cause — is that STDERR is a default stream by golang builtin log library
 
 https://golang.org/src/log/log.go#76
```go
var std = New(os.Stderr, "", LstdFlags)

// Default returns the standard logger used by the package-level output functions.
func Default() *Logger { return std }
```

kleimkuhler (approved):
@mo4islona Awesome thanks for adding the additional context; this is helpful to see. This change looks good then 👍

alpeb (approved):
LGTM :+1:

PR #37, opened 2021-04-14
Switch base image from Debian to Alpine

alpeb:
I couldn't get rid of the base image (in an attempt to get rid of
unrelated CVE warnings thrown by cluster image scanners) because of the
need of iptables and the shared libs it depends on, so the best I could
do was to switch to Alpine. Still, that reduces the image size in amd64
  from 77.3MB to 13MB.

I also changed the registry from docker hub to ghcr.io for the tester
images, because `k3d image import` was complaining (note these images
aren't pushed though).

Finally, it's no longer necessary to install `procps` (used to run
`sysctl`) as it already comes installed in Alpine.

This was tested successfully in the ARM host.

Partially addresses comment in linkerd/linkerd2#5198


Pothulapati (approved):
LGTM

dadjeibaah (approved):
📦 🚢

PR #38, opened 2021-04-20
Skip configuring firewall if rules exists

dadjeibaah:
This change fixes an issue where the `proxy-init` will fail if
`PROXY_INIT_*` chains already exist in the pod's iptables. This then
causes the pod to never start because proxy-init never finishes running
with a non-zero exit code.

In this change, we capture the output of the `iptables-save` command and
then check to see if the output contains the `PROXY_INIT_*` chains. If
they do, exist and log a warning stating that the chains already
exist.

Fixes #5786

Signed-off-by: Dennis Adjei-Baah <dennis@buoyant.io>

### DIFF
--- a/integration_test/iptables/iptablestest-lab.yaml
+++ b/integration_test/iptables/iptablestest-lab.yaml
@@ -6,25 +6,25 @@ metadata:
   labels:
     app: pod-with-no-rules
 spec:
-      containers:
-      - name: webserver
-        image: ghcr.io/linkerd/iptables-tester:v1
-        env:
-         - name: PORT
-           value: "8080"
-        command: ["go", "run", "/go/test_service/test_service.go"]
-        ports:
-        - name: http
-          containerPort: 8080
-      - name: other-container
-        image: ghcr.io/linkerd/iptables-tester:v1
-        env:
-         - name: PORT
-           value: "9090"
-        command: ["go", "run", "/go/test_service/test_service.go"]
-        ports:
-        - name: http
-          containerPort: 9090
+  containers:
+  - name: webserver
+    image: ghcr.io/linkerd/iptables-tester:v1
+    env:
+    - name: PORT
+      value: "8080"
+    command: ["go", "run", "/go/test_service/test_service.go"]
+    ports:
+    - name: http
+      containerPort: 8080
+  - name: other-container
+    image: ghcr.io/linkerd/iptables-tester:v1
+    env:
+    - name: PORT
+      value: "9090"
+    command: ["go", "run", "/go/test_service/test_service.go"]
+    ports:
+    - name: http
+      containerPort: 9090
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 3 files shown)


integration_test/iptables/iptablestest-lab.yaml
```diff
@@ -39,6 +39,71 @@ spec:
 ---
 apiVersion: v1
 kind: Pod
+metadata:
+  name: pod-with-existing-rules
+  labels:
+    app: pod-with-existing-rules
+spec:
+      containers:
```
kleimkuhler, line 47: nit: Everything under `spec` seems to be indented two extra times.

iptables/iptables.go
```diff
@@ -70,6 +75,12 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 
 	startSection("configuration")
 
+	matches := chainRegex.FindAllString(b.String(), -1)
+	if len(matches) > 0 {
+		log.Println("Found existing firewall configuration; Skipping")
+		return nil
+	}
```
kleimkuhler, line 82: So passing `-1` into `FindAllString` means we'll search the entire string for all possible matches. I would either print the matches as part of the `Println`, or just pass in `1` so that we stop at the first. There isn't really a reason to search the whole string if we're just going to check `len(matches) > 0`.

iptables/iptables.go
```diff
@@ -220,7 +231,7 @@ func makeMultiportDestinations(portsToIgnore []string) [][]string {
 	return append(destinationSlices, destinations)
 }
 
-func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd) error {
+func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd, cmdOut io.Writer) error {
```
kleimkuhler, line 234: Why is this change needed?

integration_test/iptables/iptablestest-lab.yaml
```diff
@@ -39,6 +39,71 @@ spec:
...
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            add:
+            - NET_ADMIN
+            - NET_RAW
+          privileged: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: false
+          runAsUser: 0
```
kleimkuhler, line 97: These two containers—`iptables-test` and `linkerd-init`—are supposed to conflict? So `iptables-test` I assume is created first, and then `linkerd-init` is created second which results in the rules already existing? If so, where are we confirming that? I'd expect to see something in the tests that looks at the logs for this container and asserts that.

kleimkuhler (commented):
Overall I understand the change but I have a few questions about tests and why certain changes are needed.

iptables/iptables.go
dadjeibaah: We need a way to capture the output of the current iptable rules in the pod in order to detect if they already exist. So we send pass in a buffer into string buffer which satifies the `io.Writer` interface into `executeCommand` to get that and then the caller can decide what it wants to do with that captured output.

integration_test/iptables/iptablestest-lab.yaml
dadjeibaah: Good point -- I can add that to the test to make sure we are asserting that `iptables-test` has already ran.

iptables/iptables.go
```diff
@@ -35,6 +38,7 @@ var (
 	// ExecutionTraceID provides a unique identifier for this script's execution.
 	ExecutionTraceID = strconv.Itoa(int(time.Now().Unix()))
 
+	chainRegex       = regexp.MustCompile(`-A (PROXY_INIT_OUTPUT|PROXY_INIT_REDIRECT) -m.*`)
```
alpeb, line 41: In my testing, looking at `linkerd-init`'s log I see:
```
2021/05/05 19:49:55 :; iptables-save
2021/05/05 19:49:55 # Generated by iptables-save v1.8.2 on Wed May  5 19:49:55 2021
*nat
:PREROUTING ACCEPT [0:0]
:INPUT ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
:PROXY_INIT_OUTPUT - [0:0]
:PROXY_INIT_REDIRECT - [0:0]
-A PREROUTING -m comment --comment "proxy-init/install-proxy-init-prerouting/1620244194" -j PROXY_INIT_REDIRECT
-A OUTPUT -m comment --comment "proxy-init/install-proxy-init-output/1620244194" -j PROXY_INIT_OUTPUT
-A PROXY_INIT_OUTPUT ! -d 127.0.0.1/32 -o lo -m owner --uid-owner 2102 -m comment --comment "proxy-init/redirect-non-loopback-local-traffic/1620244194" -j PROXY_INIT_REDIRECT
-A PROXY_INIT_OUTPUT -m owner --uid-owner 2102 -m comment --comment "proxy-init/ignore-proxy-user-id/1620244194" -j RETURN
-A PROXY_INIT_OUTPUT -o lo -m comment --comment "proxy-init/ignore-loopback/1620244194" -j RETURN
-A PROXY_INIT_OUTPUT -p tcp -m multiport --dports 443 -m comment --comment "proxy-init/ignore-port-443/1620244194" -j RETURN
-A PROXY_INIT_OUTPUT -p tcp -m comment --comment "proxy-init/redirect-all-outgoing-to-proxy-port/1620244194" -j REDIRECT --to-ports 4140
-A PROXY_INIT_REDIRECT -p tcp -m multiport --dports 4190,4191 -m comment --comment "proxy-init/ignore-port-4190,4191/1620244194" -j RETURN
-A PROXY_INIT_REDIRECT -p tcp -m comment --comment "proxy-init/redirect-all-incoming-to-proxy-port/1620244194" -j REDIRECT --to-ports 4143
COMMIT
```
So for `PROXY_INIT_REDIRECT` there' s more stuff before the `-m`. It might be enough to just match on the first part and drop the `-m.*`

integration_test/iptables/iptablestest-lab.yaml
```diff
@@ -39,56 +39,124 @@ spec:
...
+    - name: PORT
+      value: "9090"
+    command: ["go", "run", "/go/test_service/test_service.go"]
+    ports:
+    - name: http
+      containerPort: 9090
+  initContainers:
+  # The iptables-test init container is used to test what happens when
+  # iptable rules are run more than once. The linkerd-init container
+  # should log "Found existing firewall configuration..."
```
kleimkuhler, line 69: 👍

kleimkuhler (approved):
Looks good!

alpeb (approved):
LGTM :+1:

PR #47, opened 2021-09-23
Adds log-format for linkerd-proxy-init (#6881)

gusfcarvalho:
Linkerd-proxy-init currently did not support options to implement log formating, as opposed to other linkerd components.

This commit implements logging-format flag, to allow configure it to be a json or a plain output.

The logic applied tries to reflect the same one applied to linkerd control plane components.

This implementation changes the default logging behavior of proxy-init

Screenshot of behavior with plain format:
image

Screenshot of behavior with json format:
image

Signed-off-by: Gustavo Carvalho <gusfcarvalho@gmail.com>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,8 +2,9 @@ package cmd
 
 import (
 	"fmt"
-	"log"
 	"os/exec"
+
+	log "github.com/sirupsen/logrus"
 
 	"github.com/spf13/cobra"
 
@@ -23,6 +24,7 @@ type RootOptions struct {
 	NetNs                 string
 	UseWaitFlag           bool
 	TimeoutCloseWaitSecs  int
+	LogFormat             string
 }
 
 func newRootOptions() *RootOptions {
@@ -37,6 +39,7 @@ func newRootOptions() *RootOptions {
 		NetNs:                 "",
 		UseWaitFlag:           false,
 		TimeoutCloseWaitSecs:  0,
+		LogFormat:             "plain",
 	}
 }
 
@@ -66,6 +69,7 @@ func NewRootCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			log.SetFormatter(getFormatter(options.LogFormat))
 			return iptables.ConfigureFirewall(*config)
 		},
 	}
@@ -80,7 +84,7 @@ func NewRootCmd() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&options.NetNs, "netns", options.NetNs, "Optional network namespace in which to run the iptables commands")
 	cmd.PersistentFlags().BoolVarP(&options.UseWaitFlag, "use-wait-flag", "w", options.UseWaitFlag, "Appends the \"-w\" flag to the iptables commands")
 	cmd.PersistentFlags().IntVar(&options.TimeoutCloseWaitSecs, "timeout-close-wait-secs", options.TimeoutCloseWaitSecs, "Sets nf_conntrack_tcp_timeout_close_wait")
-
+	cmd.PersistentFlags().StringVar(&options.LogFormat, "log-format", options.LogFormat, "Configure log format ('plain' or 'json')")
 	return cmd
 }
 
@@ -113,4 +117,13 @@ func BuildFirewallConfiguration(options *RootOptions) (*iptables.FirewallConfigu
 	}
 
 	return firewallConfiguration, nil
+}
+
+func getFormatter(format string) log.Formatter {
+	switch format {
+	case "json":
+		return &log.JSONFormatter{}
+	default:
+		return &log.TextFormatter{FullTimestamp: true}
+	}
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -25,6 +25,7 @@ type RootOptions struct {
 	UseWaitFlag           bool
 	TimeoutCloseWaitSecs  int
 	LogFormat             string
+	LogLevel              string
 }
 
 func newRootOptions() *RootOptions {
@@ -40,6 +41,7 @@ func newRootOptions() *RootOptions {
 		UseWaitFlag:           false,
 		TimeoutCloseWaitSecs:  0,
 		LogFormat:             "plain",
+		LogLevel:              "info",
 	}
 }
 
@@ -59,9 +61,11 @@ func NewRootCmd() *cobra.Command {
 					fmt.Sprintf("net.netfilter.nf_conntrack_tcp_timeout_close_wait=%d", options.TimeoutCloseWaitSecs),
 				)
 				out, err := sysctl.CombinedOutput()
-				log.Println(string(out))
 				if err != nil {
+					log.Errorln(string(out))
 					return err
+				} else {
+					log.Println(string(out))
 				}
 			}
 
@@ -70,6 +74,10 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 			log.SetFormatter(getFormatter(options.LogFormat))
+			err = setLogLevel(options.LogLevel)
+			if err != nil {
+				return err
+			}
 			return iptables.ConfigureFirewall(*config)
 		},
 	}
@@ -85,6 +93,7 @@ func NewRootCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(&options.UseWaitFlag, "use-wait-flag", "w", options.UseWaitFlag, "Appends the \"-w\" flag to the iptables commands")
 	cmd.PersistentFlags().IntVar(&options.TimeoutCloseWaitSecs, "timeout-close-wait-secs", options.TimeoutCloseWaitSecs, "Sets nf_conntrack_tcp_timeout_close_wait")
 	cmd.PersistentFlags().StringVar(&options.LogFormat, "log-format", options.LogFormat, "Configure log format ('plain' or 'json')")
+	cmd.PersistentFlags().StringVar(&options.LogLevel, "log-level", options.LogLevel, "Configure log level")
 	return cmd
 }
 
@@ -126,4 +135,13 @@ func getFormatter(format string) log.Formatter {
 	default:
 		return &log.TextFormatter{FullTimestamp: true}
 	}
+}
+
+func setLogLevel(logLevel string) error {
+	level, err := log.ParseLevel(logLevel)
+	if err != nil {
+		return err
+	}
+	log.SetLevel(level)
+	return nil
 }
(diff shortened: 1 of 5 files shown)


gusfcarvalho:
So, I've opened this PR as draft because there is a point I like to discuss. Introducing logrus as the log tool opens the possibility to change logging with levels, and hence, allows the user to have more control on the outputs themselves. I'd like to know your opinions about implementing as well a log-level control, and change the log messages to be more 'compliant' regarding the new log interface proposed here.

alpeb:
Thanks for this @gusfcarvalho . It definitely makes better sense to implement logging through a properly lib like logrus instead of printf statements. Feel free to refactor that as you see fit, including getting rid of those "---" separators in the plain text output, that wouldn't make sense in the json output :wink:

gusfcarvalho:
So I guess i'm finally happy with the final result. Here are some examples when we have a config error (for instance, a container with no `NET_ADMIN` cap).
failure - debug + plain
failure - info + plain
And a normal use case example:
happy case - plain + info
happy case - json + debug

I'll open to review now =)

iptables/iptables.go
```diff
@@ -61,19 +61,18 @@ type FirewallConfiguration struct {
 // https://github.com/istio/istio/blob/e83411e/pilot/docker/prepare_proxy.sh
 func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 
-	log.Printf("Tracing this script execution as [%s]\n", ExecutionTraceID)
+	log.Debugf("Tracing this script execution as [%s]", ExecutionTraceID)
 
-	startSection("current state")
+	log.Debugln("current state")
```
mateiidavid, line 66: Since we removed the separators, I'm not sure it's worth keeping this line in. It doesn't add much to the logs and it doesn't really do anything to set the context up.

iptables/iptables.go
```diff
@@ -61,19 +61,18 @@ type FirewallConfiguration struct {
...
-		log.Println("Aborting firewall configuration")
+		log.Error("Aborting firewall configuration")
 		return err
 	}
-	endSection()
 
 	commands := make([]*exec.Cmd, 0)
 
-	startSection("configuration")
+	log.Debugln("configuration")
```
mateiidavid, line 75: Same as before, don't think this adds much.

iptables/iptables.go
```diff
@@ -85,22 +84,17 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
...
-			log.Println("Aborting firewall configuration")
+			log.Error("Aborting firewall configuration")
 			return err
 		}
 	}
 
-	endSection()
-
-	startSection("end state")
+	log.Debugln("end state")
```
mateiidavid, line 96: If we remove the `current state` we should get rid of this too

iptables/iptables.go
```diff
@@ -85,22 +84,17 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 
 	commands = addOutgoingTrafficRules(commands, firewallConfiguration)
 
-	endSection()
-
-	startSection("adding rules")
+	log.Debugln("adding rules")
```
mateiidavid, line 87: Again, don't think it makes much sense to keep this in, even at a debug level.

iptables/iptables.go
```diff
@@ -247,7 +233,7 @@ func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd,
 		cmd = exec.Command("nsenter", finalArgs...)
 	}
 
-	log.Printf(":; %s\n", strings.Trim(fmt.Sprintf("%v", cmd.Args), "[]"))
+	log.Debugf("%s", strings.Trim(fmt.Sprintf("%v", cmd.Args), "[]"))
```
mateiidavid, line 236: I think this line will print out the actual `iptables` command used, right? Personal preference for me would be to change the level to `info` so it's easier to see what commands are used. We can still leave the final `iptables-save` output as debug though.

iptables/iptables.go
```diff
@@ -61,19 +61,18 @@ type FirewallConfiguration struct {
...
 	}
-	endSection()
 
 	commands := make([]*exec.Cmd, 0)
 
-	startSection("configuration")
+	log.Debugln("configuration")
 
 	matches := chainRegex.FindAllString(b.String(), 1)
 	if len(matches) > 0 {
```
mateiidavid, line 78: Perhaps if we remove the `configuration` statement from above, we can expand a little bit here, something like:
`log.Printf("skipping iptables setup, found %d existing chains: %v", len(matches), matches)` 

or something similar to this

mateiidavid (commented):
Thank you for the PR @gusfcarvalho! This looks really good so far.

A couple of thoughts that I had after looking at attached screenshots:
  * There are a couple of lines that used to make sense before (i.e `current state`, `end state`). I think we can do away with them, my personal preference is to declutter these logs, before they made sense since the output was very verbose. We can freely change the log level now so I think they don't add much.
  * For the main iptables commands (the ones that we log out as we execute them), my preference is to keep them as `Info`. Generally, when we have questions about a packet's route through the pod, these are the most helpful lines to look at. I noted this in the comment too but I think it'd be more valuable to keep this as `Info` (so we do not have to change the log level every time we want to have a look at the rules) and then keep the actual output from `iptables-save` as debug.

cc: @alpeb and @adleong, would like to see what you think of my suggestions.

cmd/root.go
```diff
@@ -23,6 +24,8 @@ type RootOptions struct {
 	NetNs                 string
 	UseWaitFlag           bool
 	TimeoutCloseWaitSecs  int
+	LogFormat             string
```
adleong, line 27: Could we use the `ConfigureAndParse` function from "linkerd/linkerd2/pkg/flags" to configure the log level and format?

cmd/root.go
```diff
@@ -56,16 +61,23 @@ func NewRootCmd() *cobra.Command {
 					fmt.Sprintf("net.netfilter.nf_conntrack_tcp_timeout_close_wait=%d", options.TimeoutCloseWaitSecs),
 				)
 				out, err := sysctl.CombinedOutput()
-				log.Println(string(out))
 				if err != nil {
+					log.Errorln(string(out))
```
adleong, line 65: It's unclear to me why we'd sometimes log with `Errorln` and sometimes with `Error`.  I assume the logger automatically formats log messages one per line so I don't think we ever need to use function with explicitly adds newlines like this one.

cmd/root.go
```diff
@@ -56,16 +61,23 @@ func NewRootCmd() *cobra.Command {
...
 					return err
+				} else {
+					log.Println(string(out))
```
adleong, line 68: rather than using the `Print` functions (which I believe log at Info level), let's be explicit about the level and instead use something like `log.Info`.

iptables/iptables.go
adleong: I agree.  This is the most important piece of information to log so we should make sure it's printed at the default log level.

iptables/iptables.go
```diff
@@ -256,7 +242,11 @@ func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd,
 	out, err := cmd.CombinedOutput()
 
 	if len(out) > 0 {
-		log.Printf("%s\n", out)
```
adleong, line 259: Similarly, I think we should log the full output here at INFO level, regardless of whether there is an err.

iptables/iptables.go
gusfcarvalho: Ok @mateiidavid , I'm adding it on my next commit. Followup question, though, do you guys think that the chains themselves are worthy to be logged as Info level? I thought maybe putting it aside to debug one.

iptables/iptables.go
gusfcarvalho: Ok! Adding in next commit.

cmd/root.go
gusfcarvalho: My bad. Fixing it.

cmd/root.go
gusfcarvalho: @adleong I actually wanted to do that at first, however (probably due to my lack of profficiency with go), I didn't know how to properly use the existing cobra command structure at root.go and ConfigureAndParse structure itself, without having to change `main.go` and the command structure itself.
 I'll be working on it next, but if you guys have any examples you can share, I would appreciate it :smile:

adleong: Upon closer inspection, you're right that this isn't an easy fit.  `ConfigureAndParse` looks like it is mostly intended for processes which don't do their own flag parsing (like the control plane components).  Rather than try to jam it in here, I agree with keeping what you already have.

iptables/iptables.go
```diff
@@ -247,7 +228,7 @@ func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd,
 		cmd = exec.Command("nsenter", finalArgs...)
 	}
 
-	log.Printf(":; %s\n", strings.Trim(fmt.Sprintf("%v", cmd.Args), "[]"))
+	log.Infof("%s", strings.Trim(fmt.Sprintf("%v", cmd.Args), "[]"))
```
adleong, line 231: fyi this can just be `log.Info` instead of `log.Infof`.

adleong (approved):

iptables/iptables.go
mateiidavid: Yeah think that's a sensible choice.

iptables/iptables.go
```diff
@@ -111,31 +100,23 @@ func formatComment(text string) string {
...
+		log.Info("Not ignoring any uid")
 	}
 
 	// Ignore loopback
 	commands = append(commands, makeIgnoreLoopback(outputChainName, "ignore-loopback"))
 	// Ignore ports
 	commands = addRulesForIgnoredPorts(firewallConfiguration.OutboundPortsToIgnore, outputChainName, commands)
 
-	log.Printf("Redirecting all OUTPUT to %d\n", firewallConfiguration.ProxyOutgoingPort)
+	log.Printf("Redirecting all OUTPUT to %d", firewallConfiguration.ProxyOutgoingPort)
```
mateiidavid, line 119: Do we need to change this to `Infof`?

mateiidavid (approved):
Awesome! Looks good. Had two more nits but nothing that's blocking this from my side.

iptables/iptables.go
alpeb:
```suggestion
	log.Infof("Redirecting all OUTPUT to %d", firewallConfiguration.ProxyOutgoingPort)
```

PR #49, opened 2021-10-23
Linkerd without CNI as non-root

chrischdi:
Changes the image for being able to run linkerd without CNI as non-root

* adds libcap and sets capabilities on binaries required for non-root
* set USER to uid of nobody to run as non-root per default
* replace usage of `iptables-save` by `iptables -t nat -L`

New PR to continue work initially done at #31.

Credits to @pankajmt which did the initial implementation.

I set the default user to the uid of `nobody`.
I did use the uid because otherwise kubernetes requires us to set `runAsUser` explicitly because a non-numeric user is not proven to not being root.

<sup>
Christian Schlotter <christian.schlotter@daimler.com>, Daimler TSS GmbH, 
<a href="https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md">Provider Information</a>
</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,12 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -o /out/linkerd2-proxy-
 
 ## package runtime
 FROM --platform=$TARGETPLATFORM alpine:20210212
-RUN apk add iptables
+RUN apk add iptables libcap
+RUN touch /run/xtables.lock && chmod 0666 /run/xtables.lock
+RUN setcap cap_net_raw,cap_net_admin+eip /sbin/xtables-legacy-multi
 COPY LICENSE /linkerd/LICENSE
 COPY --from=golang /out/linkerd2-proxy-init /usr/local/bin/proxy-init
+RUN setcap cap_net_raw,cap_net_admin+eip /usr/local/bin/proxy-init
 ENTRYPOINT ["/usr/local/bin/proxy-init"]
+
+USER 65534
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -336,7 +336,7 @@ func makeJumpFromChainToAnotherForAllProtocols(
 }
 
 func makeShowAllRules() *exec.Cmd {
-	return exec.Command("iptables-save")
+	return exec.Command("iptables-save", "-t", "nat")
 }
 
 // asDestination formats the provided `PortRange` for output in commands.
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -56,11 +56,10 @@ type FirewallConfiguration struct {
 	UseWaitFlag            bool
 }
 
-//ConfigureFirewall configures a pod's internal iptables to redirect all desired traffic through the proxy, allowing for
+// ConfigureFirewall configures a pod's internal iptables to redirect all desired traffic through the proxy, allowing for
 // the pod to join the service mesh. A lot of this logic was based on
 // https://github.com/istio/istio/blob/e83411e/pilot/docker/prepare_proxy.sh
 func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
-
 	log.Debugf("Tracing this script execution as [%s]", ExecutionTraceID)
 
 	b := bytes.Buffer{}
@@ -94,7 +93,7 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 	return nil
 }
 
-//formatComment is used to format iptables comments in such way that it is possible to identify when the rules were added.
+// formatComment is used to format iptables comments in such way that it is possible to identify when the rules were added.
 // This helps debug when iptables has some stale rules from previous runs, something that can happen frequently on minikube.
 func formatComment(text string) string {
 	return fmt.Sprintf("proxy-init/%s/%s", text, ExecutionTraceID)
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -14,7 +14,12 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -o /out/linkerd2-proxy-
 
 ## package runtime
 FROM --platform=$TARGETPLATFORM alpine:20210212
-RUN apk add iptables
+RUN apk add iptables libcap
+RUN touch /run/xtables.lock && chmod 0666 /run/xtables.lock
+RUN setcap cap_net_raw,cap_net_admin+eip /sbin/xtables-legacy-multi
 COPY LICENSE /linkerd/LICENSE
 COPY --from=golang /out/linkerd2-proxy-init /usr/local/bin/proxy-init
+RUN setcap cap_net_raw,cap_net_admin+eip /usr/local/bin/proxy-init
```
mateiidavid, line 22: Do we need to set these on the container? I think we're setting the capabilities at a pod level already? (haven't checked if this is necessary so just curious how it is different in this case)

iptables/iptables.go
```diff
@@ -336,7 +335,7 @@ func makeJumpFromChainToAnotherForAllProtocols(
 }
 
 func makeShowAllRules() *exec.Cmd {
-	return exec.Command("iptables-save")
+	return exec.Command("iptables-save", "-t", "nat")
```
mateiidavid, line 338: > I reverted to iptables-save, however I had add -t nat as arguments to iptables-save.
 Otherwise it tries to read the wanted table from /proc/net/ip_tables_names which is only readable by root:

We might be looking at operating on the `mangle` table to add support for tproxy in the coming weeks. Do you anticipate this getting more complicated? How would our ability to add rules in different tables be affected by this change?

mateiidavid (commented):
Hey @chrischdi, thanks for working on this! I'm still going through everything that has been discussed so far. I left a few comments, mostly out of curiosity on how this change will impact our future efforts to add `tproxy` support to iptables. You can read about it here if you're curious: https://github.com/linkerd/linkerd2/issues/7089.

I'll give this a more in-depth review, my only concern is a bit superficial here: adding `-t nat` might get in the way of changing other tables. 

As far as `sysctl` goes, I think your second option sounds good, it would be great to have the ability to run the init container as root in _very specific scenarios_. Here's what I think (aside from @alpeb's comments): 
I know with sysctl we can also only configure _some_ kernel parameters that are deemed ["safe" by k8s](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#enabling-unsafe-sysctls). At the moment, we don't plan on making use of any unsafe params. I'm a bit wary of relying on `securityContext.sysctl` (and hence the first option suggested) because in the future, our requirements might change. It will be easier to make use of sysctl in code as opposed to enabling unsafe params on a per node basis (which would fall on the operator and kind of break our "works out of the box" philosophy, it would also just complicate everything).

My random thoughts aside, this looks really good!

iptables/iptables.go
chrischdi: adding rules is just as easy as `iptabels -t mangle ...` However the `makeShowAllRules` should then also include the output of `iptables-save -t mangle`.

chrischdi:
> Hey @chrischdi, thanks for working on this! I'm still going through everything that has been discussed so far. I left a few comments, mostly out of curiosity on how this change will impact our future efforts to add `tproxy` support to iptables. You can read about it here if you're curious: [linkerd/linkerd2#7089](https://github.com/linkerd/linkerd2/issues/7089).
> 
> I'll give this a more in-depth review, my only concern is a bit superficial here: adding `-t nat` might get in the way of changing other tables.

As mentioned above: the changes here should still allow adding rules to the `mangle` table via `iptables -t mangle ...`. However, it would require to concatenate the output of `iptables-save -t mangle` in the `makeShowAllRules` function to the output of `iptables-save -t nat` in this function to still see all added rules.

> As far as `sysctl` goes, I think your second option sounds good, it would be great to have the ability to run the init container as root in _very specific scenarios_. Here's what I think (aside from @alpeb's comments): I know with sysctl we can also only configure _some_ kernel parameters that are deemed ["unsafe" by k8s](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#enabling-unsafe-sysctls). At the moment, we don't plan on making use of any unsafe params. I'm a bit wary of relying on `securityContext.sysctl` (and hence the first option suggested) because in the future, our requirements might change. It will be easier to make use of sysctl in code as opposed to enabling unsafe params on a per node basis (which would fall on the operator and kind of break our "works out of the box" philosophy, it would also just complicate everything).

Setting sysctls directly from inside the container always requires the root user. This behavior is not adjustable via a container image because it is a mounted filesystem and only the root user / uid 0 has write access to the files (`sysctl -w` is just a shortcut to do a `echo "value" >> /proc/sys/my/path`) and if I'm right always requires privileged mode to be even writeable for a root user:
```
$ ls /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_close_wait -lh
-rw-r--r-- 1 root root 0 Nov  2 11:04 /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_close_wait
$ mount | grep '/proc/sys '
proc on /proc/sys type proc (ro,nosuid,nodev,noexec,relatime)
```

So setting safe or unsafe sysctls the "non-root" way would always result in using the pods `securityContext.sysctl`.

To help on the "works out of the box" philosophy: for me it would also be okay to "opt-in" for the non-root part on the linkerd chart.

So to only enable nonroot for the proxy-init container as an edge case.
For doing this the changes here would be still required.
But on the linkerd/linkerd2 part it could look like something like this:
```
{{- if .Values.proxyInit.closeWaitTimeoutSecs }}
privileged: true
runAsNonRoot: false
runAsUser: 0
{{- else }}
{{- if .Values.proxyInit.nonRoot }}
privileged: false
runAsNonRoot: true
{{- else }}
runAsNonRoot: false
runAsUser: 0
{{- end }}
```

This would keep the current behaviour, but allow everyone to opt-in to a nonRoot working proxyInit container (expect `closeWaitTimeoutSecs` is wanted.

> My random thoughts aside, this looks really good!

Thanks a lot :-)

iptables/iptables.go
mateiidavid: Awesome, sounds good!

mateiidavid (approved):
Thanks again @chrischdi! The sysctl part makes a lot of sense, thanks for confirming. From my side, this is good to 🚢 📦!

alpeb:
I'll get this merged and release v1.5.0. I think this should be launched as non-root by default, but we can discuss that in the linkerd2 PR :slightly_smiling_face:

PR #57, opened 2022-02-07
Lowercase logs and remove redundant lines

mateiidavid:
Our current logging set-up was a bit confusing when debugging. When going through each command and setting chains or rules up, we'd log what we are doing. When the commands executed, if a command failed for whatever reason, we would also get both the command and its output printed. This is a bit confusing since in practice we log all of the commands but then the failure may happen at _any_ command making it harder to pinpoint the issue.

I also lowercased everything to make it easier to grep/search.

### DIFF
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -84,7 +84,6 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 
 	for _, cmd := range commands {
 		if err := executeCommand(firewallConfiguration, cmd, nil); err != nil {
-			log.Error("aborting firewall configuration: %v", err)
 			return err
 		}
 	}
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -61,11 +61,11 @@ type FirewallConfiguration struct {
 // the pod to join the service mesh. A lot of this logic was based on
 // https://github.com/istio/istio/blob/e83411e/pilot/docker/prepare_proxy.sh
 func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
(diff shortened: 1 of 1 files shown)


iptables/iptables.go
```diff
@@ -84,7 +84,7 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 
 	for _, cmd := range commands {
 		if err := executeCommand(firewallConfiguration, cmd, nil); err != nil {
-			log.Error("Aborting firewall configuration")
+			log.Error("aborting firewall configuration: %v", err)
```
olix0r, line 87:
```suggestion
			log.Errorf("aborting firewall configuration: %v", err)
```

kleimkuhler (approved):
This looks good to me. We're basically removing each of the log lines in functions that add to `commands` because when we execute those, if one fails then we still see the actual command that failed. Makes sense to me and will make the logs easier to read.

iptables/iptables.go
```diff
@@ -84,7 +84,7 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
 
 	for _, cmd := range commands {
 		if err := executeCommand(firewallConfiguration, cmd, nil); err != nil {
-			log.Error("Aborting firewall configuration")
+			log.Errorf("aborting firewall configuration: %v", err)
```
alpeb, line 87: A similar log statement above doesn't print the error, and returns it. _If_ the caller (cobra) outputs the error, it might not be necessary to print it here.

alpeb (approved):
Looks good to me, just one nit :+1:

PR #67, opened 2022-07-12
Refactor CI pipeline

mateiidavid:
This PR refactors the set-up and build process to be more consistent 
with the tools and processed that are used in the other Linkerd repositories.

Changes made as part of this:

* Replace make with just as a build tool
* Add a workflow to lint existing actions
* Overhaul release workflow (replace outdated actions, build for multi-arch)
* Change image tags in tests to use `cr.l5d.io` instead of `ghcr`. Being
   consistent here helps with manifests that have been injected where
   only the tag needs to be changed instead of the registry.

Instead of a makefile, we now have a justfile. Instead of copying recipes,
we now have only a few recipes that aid with building or testing locally.
The recipes are used only the integration-tests workflow, since that would
mirror the set-up in CI.

We no longer build the images in the test-runner script. The test runner itself
should (imo) be separate from the build process, even if we are only building
a test image.


adleong (approved):

.github/workflows/unit_tests.yml
```diff
@@ -16,5 +16,8 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+    - uses: extractions/setup-just@v1
```
alpeb, line 19:
```suggestion
    - uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb
l```

**Review by alpeb — COMMENTED** [review]
Now I know `just`! :tada: 

> Change image tags in tests to use cr.l5d.io instead of ghcr. Being
consistent here helps with manifests that have been injected where
only the tag needs to be changed instead of the registry.

Can you clarify when is that we're needing to change that tag?
I think in the linkerd2 repo CI we tried doing things with the `ghcr.io` registry to avoid polluting the Scarf stats from CI runs (and also because Scarf was being flakey). But granted, CI runs in this repo are much less frequent.

**Comment by alpeb** [comment]
I have swapped the `Kind integration tests` required status check in the repo settings with `Proxy Init Integration Tests` so all the checks can be fulfilled :wink:

**Comment by mateiidavid** [comment]
> Now I know `just`! 🎉

Heh, it's pretty fun to use so far. Think we're going to see it shine a bit later on when we add more code to build!

> > Change image tags in tests to use cr.l5d.io instead of ghcr. Being
> > consistent here helps with manifests that have been injected where
> > only the tag needs to be changed instead of the registry.
> 
> Can you clarify when is that we're needing to change that tag? I think in the linkerd2 repo CI we tried doing things with the `ghcr.io` registry to avoid polluting the Scarf stats from CI runs (and also because Scarf was being flakey). But granted, CI runs in this repo are much less frequent.

Sure, I can expand on this. I personally find it easier to work with the images when the registry is consistent across all projects. For example, if we want to run some manual tests locally, the output from `linkerd install` will have the proxy-init images all tagged as `cr.l5d.io`, whereas here, we use `ghcr`. It's easier for me to do a quick `sed` or a `grep` for the image version and overwrite _just that_ when I want to bring in the proxy-init image I've just built.

It's not a huge simplification, it probably wouldn't take me longer to edit the whole image name, but the consistency is appealing 😅 (especially since I've started to duplicate actions/workflows across repos) . 

I might be wrong here so please correct me if that's the case but _scarf stats shouldn't change when used locally or in CI_ e.g for most CI and local runs, we'll load the image up on the k3d/kind node. With an `imagePullPolicy` of `IfNotPresent`, the kubelet will never actually download the image if it already exists on the node. In my mind _this shouldn't_ have an impact on the actual scarf stats.

File: justfile [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+  gofmt -d .
+  test -z "$(gofmt -d .)"
+
+# Run unit tests
+test-unit:
+  go test -v ./...
+
+# Run integration tests
+test-integration cluster='init-test':
+  k3d image import -c init-test {{docker_tester_tag}} {{docker_tag}}
```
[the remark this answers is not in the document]
mateiidavid: oops this is an embarrassing mistake

justfile
```diff
@@ -0,0 +1,57 @@
...
+		--platform={{arch}} \
+		--load \
+
+# Build docker image for iptables-tester (Development)
+docker-tester arch=amd64_arch:
+	docker buildx build ./integration_test \
+		--file={{dockerfile_tester_path}} \
+		--tag={{docker_tester_tag}} \
+		--platform={{arch}} \
+		--load
```
kleimkuhler, line 55: It seems indentation is not consistent with the other recipes above. Is this `just` auto formatting?

PR #68, opened 2022-07-14
Fix issue with arbitrary iptables binary

mateiidavid:
To support `nft` we have added support to use arbitrary iptables
binaries. Prior to this change, in order to use the `--wait` flag for
iptables, we'd check that the command path would be equal to "iptables".

After the change, we check if the path has "iptables" as a substring.
This has the unfortunate side-effect of applying the `-w` flag to any
command that contains the word "iptables", including "iptables-save"
which does not support the command.

This causes the container to fail when the -w flag is specified. The CNI
plugin uses the iptables library and the -w flag and this issue will
block it from working.

---

Example logs from CNI plugin:

```
/var/log # cat cni.log
time="2022-07-14T16:55:47Z" level=debug msg="linkerd-cni: cmdAdd, config parsed" prevResult="&{1.0.0 [{Name:cali8c5e64d4fbb Mac: Sandbox:}] [{Interface:<nil> Address:{IP:192.168.235.245 Mask:ffffffff} Gateway:<nil>}] [] {[]  [] []}}" version=0.3.1
time="2022-07-14T16:55:47Z" level=debug msg="linkerd-cni: setting up iptables firewall" ContainerID=be987159ed2e0865010e1a3a9071f5fe596ed6c27e2ad8a8542ba985193919b5 Namespace=linkerd Pod=linkerd-destination-58d67c5845-swrzv
time="2022-07-14T16:55:47Z" level=debug msg="linkerd-cni: adding 443 to OutboundPortsToIgnore as its a control plane component" ContainerID=be987159ed2e0865010e1a3a9071f5fe596ed6c27e2ad8a8542ba985193919b5 Namespace=linkerd Pod=linkerd-destination-58d67c5845-swrzv
time="2022-07-14T16:55:47Z" level=debug msg="tracing script execution as [1657817747]"
time="2022-07-14T16:55:47Z" level=debug msg="using 'iptables' to set-up firewall rules"
time="2022-07-14T16:55:47Z" level=debug msg="using 'iptables-save' to list all available rules"
time="2022-07-14T16:55:47Z" level=info msg="'useWaitFlag' set: iptables will wait for xtables to become available"
time="2022-07-14T16:55:47Z" level=info msg="nsenter --net=/var/run/netns/cni-9bdfd35a-53d8-86da-cd6a-44161d8197c5 -- iptables-save -t nat -w"
time="2022-07-14T16:55:47Z" level=info msg="iptables-save: unrecognized option: w\nLook at manual page `iptables-save.8' for more information.\n"
time="2022-07-14T16:55:47Z" level=error msg="aborting firewall configuration"
time="2022-07-14T16:55:47Z" level=error msg="linkerd-cni: could not configure firewall: exit status 1" ContainerID=be987159ed2e0865010e1a3a9071f5fe596ed6c27e2ad8a8542ba985193919b5 Namespace=linkerd Pod=linkerd-destination-58d67c5845-swrzv
```


olix0r (approved):

iptables/iptables.go
```diff
@@ -215,9 +215,11 @@ func makeMultiportDestinations(portsToIgnore []string) [][]string {
 }
 
 func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd, cmdOut io.Writer) error {
-	if strings.Contains(cmd.Path, "iptables") && firewallConfiguration.UseWaitFlag {
-		log.Info("'useWaitFlag' set: iptables will wait for xtables to become available")
-		cmd.Args = append(cmd.Args, "-w")
+	if firewallConfiguration.UseWaitFlag {
+		if strings.Contains(cmd.Path, "iptables") && !strings.Contains(cmd.Path, "save") {
```
adleong, line 219: Testing the binary name for various substrings seems kinda brittle and confusing.  When we call into `executeCommand`, we know whether we're doing an `iptables` type command or an `iptables-save` type command.  (I think `iptables-save` is only used for doing the show all rule?).  Can we add the `-w` flag on line 90 instead, since at that point we know that we're calling an `iptables` command.  This way, we don't need to depend on how the binary is named.

mateiidavid: Great suggestion! Moved it up to line 90. Also, we'll log whether `useWaitFlag` is enabled earlier in the execution flow so we don't log it out every time we want to exec a command.

kleimkuhler (approved):

PR #75, opened 2022-07-28
Introduce CNI validator

mateiidavid:
Pods may be initialized without any Linkerd specific routing rules when
    the CNI plugin is used. This typically tends to happen when a node is
    rebooted (or drained). Pods can be initialized on the node (host) before
    the CNI plugin has had a chance to configure the plugin list, missing
    out on iptables commands.

    This change introduces a simple container to test whether a pod has been
    configured with the required routing rules. To test for the existence of
    routing rules, the container starts a server bound on the proxy's
    outbound port (default: 4140). After the server is started, the
    container will attempt a TCP connection to the k8s api server; if
    outbound redirection rules exist in the pod's network namespace, then
    the connection should be established to the server listening on :4140.

    Under normal conditions, when rules are set-up, no outbound connections
    should be successful (since they would require passing through the
    proxy, which would not be listening when using an init container). In an
    effort to have more of a comprehensive check, we additionally generate
    some random data. The server will send the bytes to the client, if the
    bytes are received successfully, then validation passes.

    Any connection error (including timeouts), or a non-valid random
    sequence will result in the container exiting with code 111 (errno 111:
    Connection Refused). This will allow a yet-to-be-implemented controller
    to restart pods with errors.
    Move shared logic into pkg and proxy-init in own dir


.github/workflows/release.yml
```diff
@@ -59,3 +65,78 @@ jobs:
...
+    - name: Build docker images
+      shell: bash
+      run: |
+          docker buildx build . \
+          --file="cni-validator/${{ matrix.arch }}.dockerfile" \
+          --tag="ghcr.io/${{ github.repository_owner }}/cni-validator:${TAG}-${{ matrix.arch }}" \
+          --cache-from type=local,src="${DOCKER_BUILDKIT_CACHE}" \
+          --cache-to type=local,dest="${DOCKER_BUILDKIT_CACHE}",mode=max \
+          --platform="${{ matrix.arch }}" \
+          --push
```
olix0r, line 107: I have a feeling we can simplify this because we're not taking dependencies like rustls or openssl. We can use musl to build static binaries... This means we may even be able to do normal docker multiarch builds without the special arch-specific tags, etc.

olix0r: See, for example: https://github.com/olix0r/tcp-echo/blob/9e464a54e71ef6a3d5430875997eaed7048e67db/Dockerfile

olix0r:
More setup/suggestions in https://github.com/linkerd/linkerd2-proxy-init/pull/81

justfile
```diff
@@ -1,63 +1,170 @@
...
 # Run unit tests
-test-unit:
+go-test-unit:
     go test -v ./...
 
 # Run integration tests
-test-integration cluster='init-test':
+go-test-integration cluster='init-test': docker-proxy-init docker-tester
     k3d image import -c {{ cluster }} {{ docker_tester_tag }} {{ docker_tag }}
     cd integration_test && ./run_tests.sh
```
olix0r, line 128: we should pull in changes from main that decouple image loading and test running -- but I don't think the integration tests should be "go" or "rust", really. instead, they should probably be `proxy-init-test` and `validator-test`, etc.

Some of the top level recipes like `go-fmt` and `rs-fmt` probably do make sense as language scoped commands, since they operate on all sources for a language in the repo; but it's an implementation detail that the proxy-init integration tests are written in Go (and really, they're driven by a bash script and not `go test`)...

olix0r (commented):
In general, I think we should try to break this into some smaller chunks. For instance, I think we should get the code merged first with minimal tooling changes; the release process should definitely be its own PR.

With this in mind, I'm going to convert this to Draft at least until we get this synced with main.

justfile
mateiidavid: Sounds good!

mateiidavid:
> In general, I think we should try to break this into some smaller chunks. For instance, I think we should get the code merged first with minimal tooling changes; the release process should definitely be its own PR.

Will open a separate PR first to move the code in, then follow that up with another PR to update the tooling. I can tackle the release process in a third PR once everything is wired up correctly.

.github/workflows/release.yml
mateiidavid: Happy to simplify this. Curious though, we generally opt to use glibc (GNU) to dynamically link against libs like openssl to get security updates? Is there any other advantage? What makes musl builds harder to use with deps like rustls or openssl?

PR #81, opened 2022-07-30
dev: Update integration test harness

olix0r:
ustfile:

* Update the integration test docker image from Go 1.11 to 1.18
* Update the `justfile` with recipes that handle cluster creation
* Update `just test-integration` to handle building/loading dependencies
* Change `just build` to output into the `target` directory (in
  preparation for including rust in this repo).
* Use recipe dependencies to avoid unnecessarily reinvoking `just`
* Remove `DOCKER_REGISTRY` support from the justfile, as it is
  unsupported by the test runner
* Update `just docker-cache-prune` to take directory argument

Test runner:

* Rename `run_tests.sh` to `run.sh`
* Update test runner to change the directory so that it can be run from
  the root of the repo
* Do not require using the default kubectl context in test runner
* Set `-euo pipefail` in test runner; pass shellcheck
* Remove terminal manipulation from the test runner
* Use a dedicated namespace (not *default*) for test resources
* Use `kubectl wait` to wait for pods to become ready
* Use `kubectl run` instead of manually crafting a job spec
* Update the test runner to fail if the namespace already exists and
  remove the namespace if tests are successful.


mateiidavid (approved):
Looks really good, thanks for the suggestions and the clean-up!

PR #85, opened 2022-08-03
dev: Update tester to Go 1.18

olix0r:
* Update tester container to use Go 1.18
* Simplify test runner to stop matching output
* Use a fake container registry, `test.l5d.io` to prevent confusion with
  the public registries.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -7,6 +7,7 @@ on:
       - '**/*.go'
       - Dockerfile
       - go.sum
+      - integration_test/**
       - justfile
 
 jobs:
--- a/integration_test/iptables/Dockerfile-tester
+++ b/integration_test/iptables/Dockerfile-tester
@@ -1,7 +1,7 @@
-FROM golang:1.19.0
-
-ADD iptables/ /go
-# Kubernetes Jobs will be retried until they return status 0,
-# so we need to output the status for processing but make sure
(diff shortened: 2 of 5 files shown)


integration_test/iptables/Dockerfile-tester
```diff
@@ -1,7 +1,7 @@
-FROM golang:1.19.0
```
olix0r, line 1: we should update the whole repo to 1.19 or keep it all on 1.18

.github/workflows/integration.yml
```diff
@@ -7,6 +7,7 @@ on:
       - '**/*.go'
       - Dockerfile
       - go.sum
+      - integration_test/**
```
olix0r, line 10: #84 should have failed but we didn't run tests on it

integration_test/run.sh
```diff
@@ -52,7 +52,7 @@ POD_IGNORES_SUBNETS_IP=$(kip pod-ignores-subnets)
 echo "POD_IGNORES_SUBNETS_IP=${POD_IGNORES_SUBNETS_IP}"
 
 echo '# Running tester...'
-output=$(k run iptables-tester \
+k run iptables-tester \
```
olix0r, line 55: I tested that this will error if the tests fail and it does!

kleimkuhler (approved):

PR #91, opened 2022-08-05
Update Go version to v1.18

mateiidavid:
Dockerfiles for proxy-init and tester use Go v1.18 to build the source
code. This change also bumps the go.mod value. After tidying the module
file, we also get all indirect dependencies listed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,15 @@
 module github.com/linkerd/linkerd2-proxy-init
 
-go 1.16
+go 1.18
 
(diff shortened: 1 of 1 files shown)


mateiidavid:
It seems Go checks only run when `go.sum` changes. I've run the integration tests manually:

```
PASS
ok      github.com/linkerd/linkerd2-proxy-init  3.007s
```

olix0r (approved):

PR #92, opened 2022-08-05
Add validator code

mateiidavid:
This change adds the source code for a routing rule validator. Together
with the soruce code, we add the relevant Cargo manifest, a Cargo
workspace manifest, and a Dockerfile.

Additionally, the iptables shared code (i.e ports, iptables firewall
configuration) is moved into a shared pkg directory. The proxy-init
binary is moved into its own directory.


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN go mod download
 # build
 COPY . .
```
hawkw, line 11: this means we're now copying the Rust validator code into the build container for the Go `proxy-init` container. we should probably just copy the proxy-init code, so that we can cache the Go code even when the Rust code changes?

i think we might be able to just change this to
```suggestion
COPY ./proxy-init .
COPY ./pkg .
```
(unless i've overlooked some other file we also need to copy over?)

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+#[clap(version)]
+struct Args {
+    #[clap(
+        long,
+        env = "CNI_VALIDATOR_LOG_LEVEL",
+        default_value = "cni_validator=info,warn"
+    )]
+    log_level: kubert::LogFilter,
+
+    #[clap(long, default_value = "plain")]
```
hawkw, line 28: nit: other stuff also lets this be set from an env var, should we add
```suggestion
    #[clap(
        long,
        env = "CNI_VALIDATOR_LOG_FORMAT",
        default_value = "plain",
    )]
```

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+    log_format: kubert::LogFormat,
+
+    #[clap(long, default_value = "0.0.0.0:4140")]
+    outbound_proxy_addr: SocketAddr,
+
+    #[clap(parse(try_from_str), long, env = "KUBERNETES_SERVICE_HOST")]
+    target_ip: IpAddr,
+
+    #[clap(parse(try_from_str), long, env = "KUBERNETES_SERVICE_PORT")]
+    target_port: u16,
```
hawkw, line 38: tiny nit: this probably doesn't actually matter that much, since this CLI isn't actually going to be run by real humans, but it feels a tiny bit weird that we take the outbound proxy's address as an `{ip}:{port}` string, but we take the target IP and port as separate arguments? would it make more sense to make these more consistent with each other?

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+        assert_eq!(time::Duration::from_secs(120), parse_timeout("2m").unwrap());
+        assert_eq!(
+            time::Duration::from_secs(7200),
+            parse_timeout("2h").unwrap()
+        );
+        assert_eq!(
+            time::Duration::from_secs(172800),
+            parse_timeout("2d").unwrap()
+        );
+    }
```
hawkw, line 260: tiny, unimportant nit: i would just change the import to `use std::time::Duration`, and remove the `time::` prefix from these...

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+        );
+        assert_eq!(time::Duration::from_secs(120), parse_timeout("2m").unwrap());
+        assert_eq!(
+            time::Duration::from_secs(7200),
+            parse_timeout("2h").unwrap()
+        );
+        assert_eq!(
+            time::Duration::from_secs(172800),
+            parse_timeout("2d").unwrap()
+        );
```
hawkw, line 259: nit, take it or leave it: instead of `unwrap()`ping these, i might just change the assertions to 
```suggestion
        assert_eq!(Ok(time::Duration::from_secs(0)), parse_timeout("0"));
        assert_eq!(Ok(time::Duration::from_secs(0)), parse_timeout("0ms"));
        assert_eq!(Ok(time::Duration::from_secs(0)), parse_timeout("0s"));
        assert_eq!(Ok(time::Duration::from_secs(0)), parse_timeout("0m"));

        assert_eq!(
            Ok(time::Duration::from_secs(120)),
            parse_timeout("120s"),
        );
        assert_eq!(
            Ok(time::Duration::from_secs(120)),
            parse_timeout("120000ms"),
        );
        assert_eq!(Ok(time::Duration::from_secs(120)), parse_timeout("2m"));
        assert_eq!(
            Ok(time::Duration::from_secs(7200)),
            parse_timeout("2h"),
        );
        assert_eq!(
            Ok(time::Duration::from_secs(172800)),
            parse_timeout("2d"),
        );
```
that way, if any of these returns an error, the error message will be printed as part of the assertion output, so we also see the expected duration to compare against.

not a big deal though.

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+
+pub fn parse_timeout(s: &str) -> Result<time::Duration> {
+    let s = s.trim();
+    let (magnitude, unit) = if let Some(offset) = s.rfind(|c: char| c.is_ascii_digit()) {
+        let (magnitude, unit) = s.split_at(offset + 1);
+        let magnitude = magnitude.parse::<u64>()?;
+        (magnitude, unit)
+    } else {
+        anyhow::bail!("{} does not contain a timeout duration value", s);
+    };
```
hawkw, line 202: style nit, take it or leave it: this could also be
```suggestion
    let offset = s.rfind(|c: char| c.is_ascii_digit())
        .ok_or_else(|| anyhow!("{} does not contain a timeout duration value", s))?;
    let (magnitude, unit) = s.split_at(offset + 1);
    let magnitude = magnitude.parse::<u64>()?;
```
which is maybe a little simpler to read?

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+        socket
+    };
+
+    tokio::select! {
+        is_readable = stream.readable() => {
+            is_readable.context("cannot read off client socket")?;
+        }
+
+         () = tokio::time::sleep(timeout) => {
+            anyhow::bail!("timed-out ({:?}) waiting for socket to become readable", timeout);
```
hawkw, line 132: tiny nit:
```suggestion
            anyhow::bail!("timed out ({:?}) waiting for socket to become readable", timeout);
```

validator/src/main.rs
```diff
@@ -0,0 +1,261 @@
+use std::{
+    net::{IpAddr, SocketAddr},
+    process::exit,
+    sync::Arc,
+    time,
+};
+
+use anyhow::{anyhow, Context, Result};
```
hawkw, line 8: nit, take it or leave it: we use `bail!` and `ensure!` enough times here that i might import them as well? not a big deal though.

hawkw (commented):
overall, this looks good. i had some minor style suggestions, but none of them are hard blockers --- we probably don't want to copy the Rust code into the Go dockerfile, though.

also, it's probably worth adding CI checks for Rust code, now that there's Rust in this repo?

validator/src/main.rs
mateiidavid: Yeah, 100% agreed, it's super unintuitive. It's unfortunate k8s doesn't really expose an env variable that captures the whole target, and that's why the args were split. We can set the default from a different env variable here (such as `KUBERNETES_ADDR`) and configure that at the deployment level.

e.g we change to `target_addr` as an arg, and in the deployment:
```yaml
env
   - name: KUBERNETES_ADDR
      value: "$(KUBERNETES_SERVICE_HOST):$(KUBERNETES_SERVICE_PORT)"
```

validator/src/main.rs
mateiidavid: Yeah! Definitely simpler and cleaner, looks good!

validator/src/main.rs
mateiidavid: I'd rather we print the error message as part of the assertion output, so this is a great suggestion :D

mateiidavid:
Resolved all feedback, thanks again @hawkw!

> also, it's probably worth adding CI checks for Rust code, now that there's Rust in this repo?

Going to do that in a follow-up PR to keep them a bit more manageable for reviews.

validator/src/main.rs
```diff
@@ -116,7 +102,7 @@ async fn validate_outbound_redirect(
         .await
         .is_err()
     {
-        anyhow::bail!("timed-out ({:?}) waiting for server to be ready", timeout);
+        bail!("timed-out ({:?}) waiting for server to be ready", timeout);
```
hawkw, line 105: tiny nit: let's make these consistent
```suggestion
        bail!("timed out ({:?}) waiting for server to be ready", timeout);
```

hawkw (approved):
looks good to me!

validator/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+struct Args {
+    #[clap(
+        long,
+        env = "CNI_VALIDATOR_LOG_LEVEL",
+        default_value = "cni_validator=info,warn"
+    )]
+    log_level: kubert::LogFilter,
+
+    #[clap(long, env = "CNI_VALIDATOR_LOG_FORMAT", default_value = "plain")]
+    log_format: kubert::LogFormat,
```
olix0r, line 24: `CNI_` isn't really appropriate here. `LINKERD_` would be better.

also, the crate isn't called `cni_validator` so this effectively disables all non-warning logs right now

validator/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    #[clap(long, default_value = "0.0.0.0:4140")]
+    outbound_proxy_addr: SocketAddr,
+
+    #[clap(long)]
+    target_addr: SocketAddr,
+
+    #[clap(parse(try_from_str = parse_timeout), long, default_value = "120s")]
+    timeout: std::time::Duration,
```
olix0r, line 33: Is a 2m timeout really appropriate? What can change while we wait all that time? My hunch is that we want something closer to a **2s** timeout so that this program can fail quickly.

validator/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+        target_addr,
+        timeout,
+    } = Args::parse();
+
+    log_format.try_init(log_level)?;
+
+    let rng_resp = Alphanumeric.sample_string(&mut rand::thread_rng(), 8);
+    info!(%outbound_proxy_addr, %target_addr, "Validating outbound traffic is redirected to proxy's outbound port");
+
+    let (shutdown_tx, shutdown_rx) = kubert::shutdown::sigint_or_sigterm()?;
```
olix0r, line 54: I don't think that we really need graceful draining in this case. It's fine for this program to exit as soon as a signal is received. With that in mind, `tokio::signal` is probably sufficient--we don't need `kubert`/`drain` here.

validator/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+        shutdown_rx,
+        rng_resp.clone(),
+    ));
+
+    let validation_task = tokio::spawn(validate_outbound_redirect(
+        target_addr,
+        notify_ready,
+        timeout,
+        rng_resp,
+    ));
```
olix0r, line 69: I don't think we really need to have synchronization (`notify_ready`) between these tasks. Instead, it seems simpler to bind the listening address before spawning the server so that we know that any connection attempts will hit that server.

Also, I don't think there's any need to spawn the client (validation) task. We can just await on the future in the select block directly...

validator/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+            is_readable.context("cannot read off client socket")?;
+        }
+
+         () = tokio::time::sleep(timeout) => {
+            bail!("timed out ({:?}) waiting for socket to become readable", timeout);
+        }
+    };
+
+    let mut buf = [0u8; 8];
+    let read_sz = stream.read_exact(&mut buf).await?;
```
olix0r, line 127: It would be best to avoid hardcoded length in multiple places. We can either use a `const` for the length, but better, I think, is to derive it from the length of the expected token.

I don't think it matters in this case, particularly, but (I recently learned that) it's generally an anti-pattern to hold a `buf` like this across an `await`. This buf is very small, and this application isn't really performance critical, so it's not a real concern, but this data ends up held by the generated future, meaning that the executor has to hold all of this data. It's better to use a heap-allocated data structure like `Bytes` (or `Vec`, etc).

olix0r (changes requested):
All of these suggestions are covered by https://github.com/linkerd/linkerd2-proxy-init/pull/99

validator/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+ name = "validator"
```
olix0r, line 2: WDYT:

```suggestion
 name = "linkerd-network-validator"
```

validator/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    tokio::select! {
+        is_readable = stream.readable() => {
+            is_readable.context("cannot read off client socket")?;
+        }
+
+         () = tokio::time::sleep(timeout) => {
+            bail!("timed out ({:?}) waiting for socket to become readable", timeout);
+        }
+    };
```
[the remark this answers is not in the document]
mateiidavid: Yeah, I agree. I think your suggestion of using a reserved IP address works much better here. At first, I was using the k8s address. When the firewall wasn't configured, we'd establish a connection almost instantly, but then we'd timeout on the stream being readable (since the k8s api server didn't send anything to us).

Like you said it's odd the way it works, makes more sense to have a single timeout to cover everything in totality, and also it makes sense to have a smaller value for that timeout 👍🏻

validator/Cargo.toml
mateiidavid: Yeah! Sounds better, validator felt a bit vague to me :)

validator/src/main.rs
mateiidavid: Noted, will change both the name of the crate `linkerd_network_validator` and then change the env var for both log level & log format.

validator/src/main.rs
mateiidavid: 2m was completely arbitrary from my side. Lowering this to 2s sounds better.

validator/src/main.rs
```diff
@@ -0,0 +1,248 @@
...
+    debug!(client.addr = %socket.local_addr()?, "Connected");
+
+    socket
+        .readable()
+        .await
+        .context("cannot read from client socket")?;
+
+    let mut buf = BytesMut::with_capacity(size);
+    while buf.len() != size {
+        let size = socket.read_buf(&mut buf).await?;
```
mateiidavid, line 172: reposting question here since I closed the other PR:

> Will shadowing size here have an effect on the outer loop condition? Could this like finish the loop in one iteration if for example we only partially fill the buffer in the first read?

Probably not an issue here, I assume the scope inside the while loop will be different so it won't impact the condition.

validator/src/main.rs
mateiidavid: I tried to use const for length I think but then I ended up with a bunch of clones since I spawned so many tasks :( 

It makes sense for it to be an anti-pattern, I hadn't thought of it being cloned and held by the executor until you mentioned it. The usage of `Bytes` throughout the code makes more sense now, and it does simplify working with byte slices.

validator/src/main.rs
olix0r: That's right. We could change the outer `size` to `capacity` or we could change the inner size to `sz` (since it's only used in a small scope -- small scope => small name); but this would only be for readability. It's correct as-is.

Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
+members = ["linkerd-network-validator"]
```
olix0r, line 2: Not a blocker, but... I would probably leave the directory as `validator` or `network-validator` and then have the crate name be `linkerd-network-validator` -- this how we handle things in the linkerd2 repo, where the directory is `policy-controller` but the crate name is `linkerd-policy-controller`.

Basically, the `linkerd` prefix can be implied from the repo. But we want `linkerd-` in the application's name so that we can use a log setting like `linkerd=debug`, etc`. (And in general, when the binary is produced, it's better for it to be clearly marked as part of linkerd).

linkerd-network-validator/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN --mount=type=cache,target=target \
+    --mount=type=cache,from=rust:1.62.1,source=/usr/local/cargo,target=/usr/local/cargo \
+    target=$(rustup show | sed -n 's/^Default host: \(.*\)/\1/p' | sed 's/-gnu$/-musl/') ; \
+    rustup target add "${target}" && \
+    cargo build --locked --target="$target" --release --package=linkerd-network-validator && \
+    mv "target/${target}/release/linkerd-network-validator" /tmp/
+
+FROM scratch
+COPY --from=build /tmp/linkerd-network-validator /bin/
+ENTRYPOINT ["/bin/linkerd-network-validator"]
```
olix0r, line 20: Not a blocker, but do we really want a separate docker image for this?  I'd think we're going to want this to be part of the init/cni docker image.

If this isn't actually used by anything yet, I'd probably omit this file from the branch and we can sort out packaging separately.

olix0r (commented):
OK one last suggestion!

https://github.com/linkerd/linkerd2-proxy-init/pull/101 adds a basic CI workflow, removes anything related to release.

olix0r (approved):

PR #96, opened 2022-08-12
Move all go code into the `proxy-init` directory

olix0r:
In preparation for introducing components other than the `proxy-init`
binary, this change moves all of the current Go code into a
`./proxy-init` directory.

This changes the relevant Just recipes to be prefixed with
`proxy-init-`.


mateiidavid (approved):
Looks good, I do have a question around the code structure.

We anticipate moving the CNI plugin into this repository at some point. The CNI plugin will make use of the shared firewall config lib, would it make sense to keep the `pkg/` directory in the root as opposed to having it in `proxy-init`? That way we could import them directly for the CNI plugin? This might be possible with [`go workspaces`](https://pkg.go.dev/cmd/go#hdr-Workspace_maintenance) which were introduced in go 1.18.

olix0r:
> We anticipate moving the CNI plugin into this repository at some point. The CNI plugin will make use of the shared firewall config lib, would it make sense to keep the pkg/ directory in the root as opposed to having it in proxy-init? That way we could import them directly for the CNI plugin? This might be possible with [go workspaces](https://pkg.go.dev/cmd/go#hdr-Workspace_maintenance) which were introduced in go 1.18.

FWIW, the CNI plugin _also_ has dependencies on the proxy-init/cmd module:
```
#; linkerd2@ver/ci-nextest-fixup#b3b528000
:; rg proxy-init/ 
./controller/api/destination/watcher/opaque_ports_watcher.go
7:      "github.com/linkerd/linkerd2-proxy-init/ports"

./cni-plugin/main.go
33:     "github.com/linkerd/linkerd2-proxy-init/cmd"
34:     "github.com/linkerd/linkerd2-proxy-init/iptables"

./pkg/util/parsing.go
7:      "github.com/linkerd/linkerd2-proxy-init/ports"
```

The cni-plugin **also** takes dependencies more deeply from the linkerd2 repo:
```
	"github.com/linkerd/linkerd2/pkg/k8s"
```

I think we're going to want to drop this dependency, or we're going to want to move the portrange code to the Linkerd2 repo so we want can avoid having circular dependencies between these repos. FWIW, the linkerd2 repo is really not setup today to be used as a library from other repos--we don't do proper semver, at the very least. I'd advocate for updating 

I guess `pkg` is fine for library code... We use it elsewhere, but we obviously haven't been very consistent about it. It's somewhat unfortunate that we need to move it at all, since this is a "breaking" change and should cause a major version bump in this repo... I guess if we're going to break the API, it would be better to do so once and not have to do it over and over...

* We rename `ports` to `portrange` (for clarity).
* We duplicate the `portrange` code in the Linkerd2 repo rather than taking a dependency. It's about 50 lines total. Who cares. It's not complicated (though I did just find a bug).
* We move the iptables and portrange modules into proxy-init -- we explicitly do **not** expose library code from this repo. We drop the `pkg/` prefix (since it's used to connote
* cni-plugin will have to take dependencies on proxy-init code, or we unify the binaries with subcommands or...
* We bump the major version of this repo :( And we do **not** update Linkerd2 to use the new version.

kleimkuhler (approved):

PR #99, opened 2022-08-16
Suggestions for #92

olix0r:
* There's no need to have synchronization between the client and server.
  We can instead bind the listener before doing anything with the client
  to ensure that the process is listening.
* Use a longer token string, ended with a newline (for aesthetics when
  using netcat, etc).
* Use `bytes` to store the server token.
* Change environment prefix to `LINKERD_VALIDATOR_` (instead of `CNI_`).
* Change the default log level to simply `info`. (No logs were emitted
  by default).
* Rename `target-addr` to `connect-addr` and rename
  `outbound-proxy-addr` to `server-addr`. This makes it easier to
  understand them in the context of this application.
* Apply a single top-level timeout, instead of applying the timeout
  independently on each call.
* Change the timeout to 10s. There's no use waiting for 2 minutes when
  we don't do any sort of retrying, etc.
* Replace `kubert::shutdown` with `tokio::signal` -- we don't actually
  need graceful of draining of anything in the process, we just need to
  stop the process when a signal is received.
* Add comments throughout.

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -13,7 +13,7 @@ use tracing::{debug, error, info, Instrument};
 /// Validate that a container's networking is setup for the Linkerd proxy
 ///
 /// Validation is done by binding a server on the proxy's outbound port and
-/// initiatinga connection to an arbitrary (hopefully unroutable) address. If
+/// initiating a connection to an arbitrary (hopefully unroutable) address. If
 /// networking has been configured properly, the connection should be
 /// established to the server.
 #[derive(Parser)]
@@ -90,14 +90,14 @@ async fn main() {
     }
 }
 
-// === Valdiation ===
+// === Validation ===
 
 /// Validates that connecting to `connect_addr` actually connects to
 /// `listen_addr`.
 ///
-/// This validates the the operating system (i.e. iptables) is configured to
+/// This validates that the operating system (i.e. iptables) is configured to
 /// redirect connections to a Linkerd proxy (with an outbound port of
-/// `listen_addr).
+/// `listen_addr`).
 async fn validate(listen_addr: SocketAddr, connect_addr: SocketAddr) -> Result<()> {
     // First, bind the server address so that all connections can be processed
     // by the server.
@@ -118,7 +118,7 @@ async fn validate(listen_addr: SocketAddr, connect_addr: SocketAddr) -> Result<(
     tokio::spawn(serve(listener, token.clone()).in_current_span());
 
     // Connect to an arbitrary address, read data from the connection, and fail
-    // if it does match the server's token.
+    // if it doesn't match the server's token.
     info!("Connecting to {connect_addr}");
     let data = connect(connect_addr, token.len()).await?;
     debug!(data = ?String::from_utf8_lossy(&*data), size = data.len());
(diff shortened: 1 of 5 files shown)


validator/src/main.rs
```diff
@@ -1,186 +1,182 @@
...
         }
-    };
+    }
+}
 
-    let mut buf = [0u8; 8];
-    let read_sz = stream.read_exact(&mut buf).await?;
-    let resp = String::from_utf8(buf[..expected_resp.len()].to_vec())?;
-    debug!(redirect_response = %resp, bytes_read = %read_sz);
+// === Valdiation ===
```
mateiidavid, line 93:
```suggestion
// === Validation ===
```

validator/src/main.rs
```diff
@@ -1,186 +1,182 @@
...
+
+/// Validates that connecting to `connect_addr` actually connects to
+/// `listen_addr`.
+///
+/// This validates the the operating system (i.e. iptables) is configured to
```
mateiidavid, line 98:
```suggestion
/// This validates that the operating system (i.e. iptables) is configured to
```

validator/src/main.rs
```diff
@@ -1,186 +1,182 @@
...
+/// redirect connections to a Linkerd proxy (with an outbound port of
+/// `listen_addr).
```
mateiidavid, line 100:
```suggestion
/// `listen_addr`).
```

validator/src/main.rs
```diff
@@ -1,186 +1,182 @@
...
+    );
+    debug!(?token);
+    let token = Bytes::from(token);
+
+    // Spawn a server on a background task that writes the response and then
+    // closes client the connection.
+    tokio::spawn(serve(listener, token.clone()).in_current_span());
+
+    // Connect to an arbitrary address, read data from the connection, and fail
+    // if it does match the server's token.
```
mateiidavid, line 121:
```suggestion
    // if it doesn't match the server's token.
```

mateiidavid (approved):
Great suggestions, looks cleaner :)

validator/src/main.rs
```diff
@@ -1,186 +1,182 @@
...
+    debug!(client.addr = %socket.local_addr()?, "Connected");
+
+    socket
+        .readable()
+        .await
+        .context("cannot read from client socket")?;
+
+    let mut buf = BytesMut::with_capacity(size);
+    while buf.len() != size {
+        let size = socket.read_buf(&mut buf).await?;
```
olix0r, line 172: we need to detect `size == 0` here....

olix0r:
@mateiidavid i'll let you merge this into your branch...

validator/src/main.rs
```diff
@@ -1,186 +1,185 @@
...
```
mateiidavid, line 172: Will shadowing `size` here have an effect on the outer loop condition? Could this like finish the loop in one iteration if for example we only partially fill the buffer in the first read?

olix0r: Practically there are no issues here, but it may be confusing enough to a reader that we might want to call this variable `sz`...

> Could this like finish the loop in one iteration if for example we only partially fill the buffer in the first read?

If we only partially fill the buffer, read_buf will return the number of bytes read and `size` will reflect that value.

The outer `size` is **only** used as the capacity for the buffer in this code.

PR #115, opened 2022-10-10
release: Split proxy-init and validator releases

olix0r:
* Tags in the form `proxy-init/v2.0.1` trigger proxy-init releases with docker artifacts.
* Tags in the form `validator/v0.1.0` trigger validator releases with static binary artifacts.
* The release workflows are now tested when the workflow changes.


.github/workflows/release-proxy-init.yml
```diff
@@ -0,0 +1,96 @@
...
+            echo ::set-output name=mode::release
+          else
+            sha='${{ github.sha }}'
+            echo ::set-output name=version::"test-${sha:0:7}"
+            echo ::set-output name=mode::test
+          fi
+    outputs:
+      repo: ghcr.io/${{ github.repository_owner }}/proxy-init
+      mode: ${{ steps.meta.outputs.mode }}
+      version: ${{ steps.meta.outputs.version }}
```
hawkw, line 36: nit, take it or leave it: would it be worth factoring this out into a bash script/action that can just be run with the expected regex? not a big deal...

hawkw (approved):
looks good to me!

stevej (approved):
LGTM. Thanks for doing this.

.github/workflows/release-proxy-init.yml
```diff
@@ -0,0 +1,96 @@
+name: Proxy-init release
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/release-proxy-init.yml
```
kleimkuhler, line 6: All jobs will run when this file changes, right? Do we want that? We don't do this in `release.yml` for linkerd2.

If we do want this to run to make sure it still works properly, should we make some of the jobs conditional on it being a tag that was pushed?

PR #125, opened 2022-10-18
init -> proxy-init

stevej:
Small fix to proxy-init release pipeline.

The release workflow in jobs.meta looks for tags matching 'proxy-init' while the workflow in on.push looks for tags matching 'init'. This change makes the two patterns match.

I pushed a tag called `init/v2.1.0` and noticed it was triggering the test workflow. This should correct that.


mateiidavid (approved):
Nice catch!

PR #147, opened 2022-12-05
PoC spike for k8s-based linkerd-cni testing

stevej:
PoC spike for testing integration of linkerd-cni within k3s. The initial test uses flannel as that's what k3s uses by default.

This branch is intended to kick off a discussion of ways to perform integration testing of linkerd-cni with commonly shared CNIs. Do we like this approach?

* Github workflow rules for running linkerd-cni integration tests.
* Dockerfile for linkerd-cni integration tests
* Integration tests that ensure that linkerd-cni's `install-cni.sh` interacted nicely with flannel's `conf` file.
* `linkerd-cni.yaml` generated from `linkerd install-cni` with arguments for running within k3d
* `run.sh` executes the containerized integration tests within k8s.
* justfile rules for building linkerd-cni Dockerfile and cni-plugin-test Dockerfile for integration tests
* justfile rules for running integration tests for linkerd-cni


justfile
```diff
@@ -130,12 +137,30 @@ build-cni-plugin-image *args='--load':
...
+# Run integration tests without preparing dependencies
+cni-plugin-test-integration-run:
+    TEST_CTX="k3d-$(just-k3d --evaluate K3D_CLUSTER_NAME)" ./cni-plugin/integration/run.sh
+
+
+
+# TODO(stevej): add a k3d-create-debug
 export K3S_DISABLE := "local-storage,traefik,servicelb,metrics-server@server:*"
-export K3D_CREATE_FLAGS := '--no-lb'
+export K3D_CREATE_FLAGS := '--no-lb --k3s-arg "--debug@server:*"'
```
stevej, line 163: I need to move this into a flag for creating a cluster with debug logging.

mateiidavid: Hmmm, I have in mind perhaps two options. Both of them involve creating a cluster with `--debug` by default. I think it makes sense to have it on my default and then override it in CI (if we want to). Maybe this particular way of doing things wouldn't be a good fit for the other linkerd repos but in this case I think it makes sense, especially since we are going to be pretty reliant on kubelet logs for debugging. Anyway:

1) We could use something like: 

```
_default_k3d_flags := '--no-lb --k3s-arg "--debug@server:*"'
export K3D_CREATE_FLAGS := env_var_or_default("K3D_CREATE_FLAGS", _default_k3d_flags)
```

One of way of doing this is to export the environment variable based on a recipe-local temp var. Sounds like a mouthful but it means that by default we will use `--debug`. We can then override this either locally:

```
$ just _default_k3d_flags='--no-lb' k3d-create
```

or in CI using the same syntax.

Another approach here is to simply hardcode the default value in `K3D_CREATE_FLAGS`, locally and CI we can always override by setting the env variable.

2) We could have a recipe that handles this for us, e.g `_k3d_create_init` or something similar. We'd make use of a variable again to store default config and then export K3D_CREATE_FLAGS to the rest of the environment. This is more in line with what you're suggesting, but I think it just adds a bit more indirection when we can solve it maybe in a simpler way if we use `env_var_or_default`.

I'm by no means a just or makefile expert tho so lmk what you think.

cni-plugin/integration/flannel/flannel_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"os"
+	"testing"
+)
+
+const (
+	ConfigDirectory = "/var/lib/rancher/k3s/agent/etc/cni/net.d"
+	FlannelConflist = "10-flannel.conflist"
+)
+
+func ls(dir string, t *testing.T) []string {
```
mateiidavid, line 16: As an alternative, we could return an error instead of relying on `t`. This would maybe make it a bit clearer that the function is purely a utility, but it does mean we will end up handling more error cases in our tests (copy&pasta ftw).

Not bothered either way, will leave it up to you :D

cni-plugin/integration/flannel/flannel_test.go
```diff
@@ -0,0 +1,93 @@
...
+	files, err := os.ReadDir(dir)
+	if err != nil {
+		t.Fatalf("failed to list files: %v", err)
+	}
+	fileNames := make([]string, len(files))
+	for i, f := range files {
+		fileNames[i] = f.Name()
+	}
+	return fileNames
```
mateiidavid, line 25: Ok...unpopular opinion time, if we do something unorthodox here and store the names as keys in a map we can avoid `contains` and we can simply do one-liners throughout the code 

```go
if _, ok := mymap[FlannelConflist]; !ok {
  t.Fatalf("%s not found", FlannelConflist)
}
``` 

We can limit the amount of space we use by storing `struct{}` ([which has 0 bytes in sz!](https://dave.cheney.net/2014/03/25/the-empty-struct)):

```go
fileNames := make(map[string]struct{}, len(files))
```

I can understand if you do not like this approach though and will not take it personal 😂

cni-plugin/integration/flannel/flannel_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if !*runTests {
+		fmt.Fprintln(os.Stderr, "integration tests not enabled: enable with -integration-tests")
+		os.Exit(0)
+	}
+
+	os.Exit(m.Run())
+}
+
+func TestLinkerdCNIIsLastPlugin(t *testing.T) {
```
mateiidavid, line 50: Super nitpicky but CNIIS might be a bit hard to parse

```suggestion
func TestLinkerdIsLastCNIPlugin(t *testing.T)
```

cni-plugin/integration/run.sh
```diff
@@ -0,0 +1,98 @@
...
+    kubectl "$@"
+  fi
+}
+
+function create_test_lab() {
+    echo '# Creating the test lab...'
+    k create ns cni-plugin-test
+    k create serviceaccount linkerd-cni
+    # TODO(stevej): how can we parameterize this manifest with `version` so we
+    # can enable a testing matrix?
```
mateiidavid, line 21: Something I've done to template out a node-debugger pod manifest whenever I needed to change `nodeName` before applying is to replace the value with a placeholder and then use sed.

```
nodeName: <replace-me>
```

`cat manifest.yaml | sed -e 's/<replace-me>/node-foo/g` | kubectl apply -f -`

I'm not saying it's a good solution but it might do until we move -- if we ever decide to -- to a harness that programatically creates two resources.

cni-plugin/integration/run.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# This needs to use the name linkerd-proxy so that linkerd-cni will run.
+echo '# Running tester...'
+k run linkerd-proxy \
+        --attach \
+        --image="test.l5d.io/linkerd/cni-plugin-tester:test" \
+        --image-pull-policy=Never \
+        --namespace=cni-plugin-test \
+        --restart=Never \
+        --overrides='{
```
mateiidavid, line 71: neat had no idea you can use `--overrides`

mateiidavid (commented):
Looks really good 👌🏻 tbh there's nothing really blocking this from my side. I've left a few comments and will check back in but regardless of the outcome, as a first iteration I think this is good.

cni-plugin/integration/flannel/flannel_test.go
stevej: Using a map with empty keys here is a good idea, better than my 1970's approach for sure. I'll change it and also change these functions to return err instead of passing in a `*testing.T`

cni-plugin/integration/flannel/Dockerfile-tester
```diff
@@ -0,0 +1,10 @@
+# syntax=docker/dockerfile:1.4
+
```
stevej, line 2: TODO: document why this doesn't have an ENTRYPOINT

cni-plugin/integration/manifests/linkerd-cni.yaml
```diff
@@ -0,0 +1,177 @@
+##
+## Everything below here is generated from the output `linkerd install-cni`
+## and modified with the test image of the cni-plugin.
```
stevej, line 3: It actually isn't the test image. It will be the test image when we're done.

cni-plugin/integration/run.sh
```diff
@@ -0,0 +1,98 @@
...
+    --attach \
+    -i \
+    --command \
+    --image="cr.l5d.io/linkerd/proxy:edge-22.12.1" \
+    --image-pull-policy=IfNotPresent \
+    --namespace=cni-plugin-test \
+    --restart=Never \
+    --rm \
+    -- \
+    /usr/lib/linkerd/linkerd2-network-validator --log-format plain --log-level debug --connect-addr 1.1.1.1:20001 --listen-addr 0.0.0.0:4140 --timeout 10s
```
stevej, line 60: break up this line with \

cni-plugin/integration/run.sh
stevej: Now that you mention it, we could use `sed` to replace one image with another image.

cni-plugin/integration/run.sh
stevej: One of my attempts was to use ovrrides with an initContainers but that doesn't work and silently fails as you can't put an init container on a container that's already running.

cni-plugin/integration/run.sh
```diff
@@ -0,0 +1,105 @@
...
+    -- \
+    /usr/lib/linkerd/linkerd2-network-validator --log-format plain \
+      --log-level debug --connect-addr 1.1.1.1:20001 \
+      --listen-addr 0.0.0.0:4140 --timeout 10s
+
+echo 'PASS: Network Validator'
+
+# the integration tests to run. pass in as an environment variable.
+# defaults to the tests in the flannel subdirectory
+SUBDIRECTORY=${SUBDIRECTORY-flannel}
```
stevej, line 68: rename this to SCENARIO

mateiidavid (approved):
Looks good!

cni-plugin/integration/run.sh
```diff
@@ -0,0 +1,105 @@
...
+if k get ns/cni-plugin-test >/dev/null 2>&1 ; then
+  echo 'ns/cni-plugin-test already exists' >&2
+  exit 1
+fi
+
+create_test_lab
+
+# Wait for linkerd-cni daemonset to complete
+if ! k rollout status --timeout=30s daemonset/linkerd-cni -n linkerd-cni; then
+  echo "!! linkerd-cni didn't rollout properly, check logs";
```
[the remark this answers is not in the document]
stevej: I've added this and I've also taken the opportunity to beef up error handling in the `cleanup` function. If one of those functions errored then the rest of the cleanup wouldn't happen properly and that was annoying.

justfile
stevej: I turned this into an issue in the kanban board, number 65 there.

cni-plugin/integration/run.sh
```diff
@@ -0,0 +1,107 @@
...
+  exit 1
+fi
+
+create_test_lab
+
+# Wait for linkerd-cni daemonset to complete
+if ! k rollout status --timeout=30s daemonset/linkerd-cni -n linkerd-cni; then
+  echo "!! linkerd-cni didn't rollout properly, printing logs";
+  k describe ds linkerd-cni
+  k logs linkerd-cni -n linkerd-cni
```
mateiidavid, line 45: we don't need a printf || echo here, do we?

mateiidavid (approved):
Still looks good :D

cni-plugin/integration/run.sh
stevej: Not strictly since if line 44 fails, line 45 would almost certainly fail but I think I'd like to err on having more information so I've put `|| echo` on these lines as well.

PR #148, opened 2022-12-05
Replace pkg/ with internal/

olix0r:
There's no need for a public library export. We can share code within this repo via the `internal` directory.


proxy-init/cmd/root.go
```diff
@@ -7,8 +7,8 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
-	"github.com/linkerd/linkerd2-proxy-init/pkg/linkerd-iptables"
-	"github.com/linkerd/linkerd2-proxy-init/pkg/linkerd-util"
+	iptables "github.com/linkerd/linkerd2-proxy-init/internal/linkerd-iptables"
+	util "github.com/linkerd/linkerd2-proxy-init/internal/linkerd-util"
```
olix0r, line 11: We could probably stop prefixing the package names with linkerd- now that they are internal.

stevej (approved):
That's a great idea.

PR #161, opened 2023-01-05
Dockerfile for cni-plugin and docker-based tests for CNI installer script

stevej:
* justfile workflow for running docker-based integration tests
* Dockerfile-cni-plugin for building those tests along with the cni-plugin
* github workflow for running those tests when anything touches the installer
* Updated a comment in `install-cni.sh` to trigger the installer integration test

### DIFF
--- a/.github/workflows/cni-plugin-installer.yml
+++ b/.github/workflows/cni-plugin-installer.yml
@@ -1,0 +1,19 @@
+name: cni-plugin-installer
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - cni-plugin/deployment/scripts/**
+
+jobs:
+  run-tests:
+    timeout-minutes: 15
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: linkerd/dev/actions/setup-tools@v38
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+      - uses: extractions/setup-just@95b912dc5d3ed106a72907f2f9b91e76d60bdb76
+      - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18
+      - uses: docker/setup-buildx-action@8c0edbc76e98fa90f69d9a2c020dcb50019dc325
+      - run: just cni-plugin-installer-integration-run
--- a/.github/workflows/cni-plugin-installer.yml
+++ b/.github/workflows/cni-plugin-installer.yml
@@ -14,6 +14,5 @@ jobs:
       - uses: linkerd/dev/actions/setup-tools@v38
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
       - uses: extractions/setup-just@95b912dc5d3ed106a72907f2f9b91e76d60bdb76
-      - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18
       - uses: docker/setup-buildx-action@8c0edbc76e98fa90f69d9a2c020dcb50019dc325
       - run: just cni-plugin-installer-integration-run
--- a/Dockerfile-cni-plugin
+++ b/Dockerfile-cni-plugin
@@ -1,10 +1,12 @@
 # syntax=docker/dockerfile:1.4
 
-ARG BUILDPLATFORM=linux/amd64
+##
+## Go
+##
 
-# Precompile key slow-to-build dependencies
-FROM --platform=$BUILDPLATFORM golang:1.18-alpine as golang
-WORKDIR /linkerd-build
+# Cross compile from native platform to target arch
+FROM --platform=$BUILDPLATFORM ghcr.io/linkerd/dev:v38-go as go
+WORKDIR /build
 COPY --link go.mod go.sum .
 COPY --link ./cni-plugin ./cni-plugin
 COPY --link ./proxy-init ./proxy-init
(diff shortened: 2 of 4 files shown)


Dockerfile-cni-plugin
```diff
@@ -0,0 +1,40 @@
+# syntax=docker/dockerfile:1.4
+
+ARG BUILDPLATFORM=linux/amd64
+
+# Precompile key slow-to-build dependencies
+FROM --platform=$BUILDPLATFORM golang:1.18-alpine as golang
+WORKDIR /linkerd-build
+COPY --link go.mod go.sum .
```
mateiidavid, line 8: TIL of `--link`. If I understand it, `--link` basically produces its own layer which at the end is linked together with the rest of the dockerfile. Without it, we'd depend on the previous layer.

This means that caching should work better since we won't need to compute any diffs, or we won't need to compute the previous layers? Not sure my question makes sense

Dockerfile-cni-plugin
```diff
@@ -0,0 +1,40 @@
...
+COPY --link ./cni-plugin ./cni-plugin
+COPY --link ./proxy-init ./proxy-init
```
mateiidavid, line 10: Welll... probably not the best place to put this comment but we are building and using `proxy-init` only to use the `RootCommand` struct. We could perhaps  build `iptables.FirewallConfiguration` ourselves, or find another way to pass the options in so we can avoid coupling the CNI plugin and proxy-init together (outside of their shared `internal` package).

Dockerfile-cni-plugin
```diff
@@ -0,0 +1,40 @@
+# syntax=docker/dockerfile:1.4
+
+ARG BUILDPLATFORM=linux/amd64
```
mateiidavid, line 3: Does this have an effect on whether building happens strictly for `amd64`? i.e if we wanna build this for arm, will overriding this variable have an effect?  afaik the value is taken from the environment when docker is run.

Dockerfile-cni-plugin
```diff
@@ -0,0 +1,40 @@
...
+
+# Precompile key slow-to-build dependencies
+FROM --platform=$BUILDPLATFORM golang:1.18-alpine as golang
```
mateiidavid, line 6: could use

```
ghcr.io/linkerd/dev:v38-go as go
```

instead of golang:1.18-alpine?

Dockerfile-cni-plugin
```diff
@@ -0,0 +1,40 @@
...
+    jq && \
+    rm -rf /var/lib/apt/lists/*
+
+# We still rely on old iptables-legacy syntax.
+RUN update-alternatives --set iptables /usr/sbin/iptables-legacy \
+    && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+
+COPY --from=golang /go/bin/linkerd-cni /opt/cni/bin/
+COPY LICENSE .
+COPY cni-plugin/deployment/scripts/install-cni.sh .
```
mateiidavid, line 36: do we need to specify:

```
--from=go
```

in these copies so we can copy from the Go container? Or does this take the files from the build context? Bit of a noob question on my end.

mateiidavid (commented):
Perused the dockerfile a little bit. This looks ready to ship to me. My docker's a bit rusty (especially around multi-arch builds) so I thought I'd ask some questions here before diving into any independent research :D.

Dockerfile-cni-plugin
stevej: I think this was a mistake, I've taken `BUILDPLATFORM` out. An explicit goal here is to essentially duplicate what the `proxy-init` Dockerfile is doing and it's not doing this.

Dockerfile-cni-plugin
stevej: We also use `BuildFirewallCommand` but I agree with you that can be pulled out since `BuildFirewallCommand` returns and `*iptables.FirewallConfiguration`. What do you think about putting it in a card and we can tackle it in a few weeks?

Dockerfile-cni-plugin
stevej: That's my understanding as well. I'm following what @olix0r has done in the `proxy-init` Dockerfile so I'm going to invite him to chime in here.

Dockerfile-cni-plugin
stevej: I'm pretty sure this works but I went ahead and changed it to `--from=go` to ensure we're doing it uniformly and maybe it saves us a few megabytes in the image size.

Dockerfile-cni-plugin
mateiidavid: A card sounds good 👌🏻

Dockerfile-cni-plugin
mateiidavid: I can research it a little bit and post my findings somewhere 👍🏻

Dockerfile-cni-plugin
mateiidavid: Seems like the change is now breaking the docker build. Perhaps we can remove the `--from` for now. No idea why it can't compute the cache key when `--from` is specified. Hm

.github/workflows/cni-plugin-installer.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  run-tests:
+    timeout-minutes: 15
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: linkerd/dev/actions/setup-tools@v38
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+      - uses: extractions/setup-just@95b912dc5d3ed106a72907f2f9b91e76d60bdb76
+      - uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18
```
mateiidavid, line 17: Might not need this. We can just run the tests using the amd64 build.

mateiidavid (approved):
Looks good so far. It's good to ship if needed for the poc 👌🏻

PR #164, opened 2023-01-06
Make iptables rule application idempotent upon init pod restart

jim-minter:
Fixes https://github.com/linkerd/linkerd2/issues/9267

If the init pod restarts for any reason when the iptables rules are half-applied, currently we skip re-applying them, leaving things in a broken state.

Instead, behave idempotently.  The first run of the init pod is unchanged by this PR.  However if there is a second run, it will flush any pre-existing chains and re-add the rules, and it will only add the rules that jump to the chains if they are not found.

Signed-off-by: Jim Minter <jimminter@microsoft.com>


jim-minter:
cc @kleimkuhler @mattstam

jim-minter:
@linkerd/maintainers, please can I have a review on this?

mattstam (approved):

internal/iptables/iptables.go
```diff
@@ -270,12 +263,16 @@ func (fc FirewallConfiguration) makeIgnoreUserID(chainName string, uid int, comm
...
+func (fc FirewallConfiguration) makeFlushChain(name string) *exec.Cmd {
 	return exec.Command(fc.BinPath,
 		"-t", "nat",
-		"-N", name,
-		"-m", "comment",
-		"--comment", formatComment(comment))
+		"-F", name)
+}
+
+func (fc FirewallConfiguration) makeCreateNewChain(name string) *exec.Cmd {
```
stevej, line 272: Would you please add a comment flag here?

jim-minter: I removed it because as far as I can see, although iptables accepts `-m comment --comment` with the `-N` command, it doesn't actually do anything with a comment specified on a chain.  For example, running `iptables-save` doesn't show the comment subsequently.  Also `man iptables-extensions` states "comment: Allows you to add comments ... to any **rule**" (emphasis mine).

iptables is really picky about matching the exact specification of a rule when deleting it - i.e. if you run `-A foo -m comment --comment bar`, you must run `-D foo -m comment --comment bar`, but this is not the case with chains.

Because of this difference in behavior, I thought it was clearer to omit the comment on `-N`, which is unhonored anyway, just as there is no logic to try to retrieve the comment previously used when running `-F`.

Please reconfirm if you want the comment parameter readded?

internal/iptables/iptables.go
```diff
@@ -126,32 +124,40 @@ func (fc FirewallConfiguration) addOutgoingTrafficRules(commands []*exec.Cmd) []
...
+	if outputRuleRegex.Find(existingRules) == nil {
+		// Redirect all remaining outbound traffic to the proxy.
+		commands = append(
+			commands,
+			fc.makeJumpFromChainToAnotherForAllProtocols(
+				IptablesOutputChainName,
+				outputChainName,
+				"install-proxy-init-output",
+				false))
+	}
```
kleimkuhler, line 136: So my main question here and also in `addIncomingTrafficRules` is why the `nil` check is necessary. Maybe I'm missing something with creating or flushing the chain.

At the start of both of these functions, we either call `makeCreateNewChain` or `makeFlushChain`; either function we call would expect the rules for this chain to be empty. So, when would we expect `outputRuleRegex.Find(existingRules) != nil`? I think with the checks that you added above, we can always expect this to be `nil`. If that is a wrong assumption then please correct me!

kleimkuhler (commented):
Overall this is looking good. I have a question about a `nil` check that has been added to both of the outgoing/incoming functions and why it's necessary. Thanks for working on this though and sorry for the late review!

internal/iptables/iptables.go
jim-minter: This module modifies four chains.  Two are PROXY_INIT_OUTPUT and PROXY_INIT_REDIRECT, which we own outright.  These we can either create or flush knowing that all the rules inside belong to us.  We decide what to do based on regexes outputChainRegex and redirectChainRegex respectively.

The other two are OUTPUT and PREROUTING, where we want to ensure one rule that jumps to the aforementioned chains.  I'm being more careful here because it's possible that there could be other user-owned rules in these system chains.  We shouldn't just flush and repopulate them because we could accidentally remove other user-owned rules.  So we detect the existence of the one rule we want to ensure using regexes outputRuleRegex and preroutingRuleRegex respectively.

We need to check what to do independently in each of these four cases because of the possibility that we crashed previously half way through the configuration.  This explains why there are 4 `regex.Find() == nil` checks - each uses a different regex and decides what to do with a different chain.

jim-minter:
Hey @stevej, @kleimkuhler, do you have any more thoughts on this?

internal/iptables/iptables.go
kleimkuhler: Thanks for the explanation. Because of the fact that we may not own the OUTPUT and PREROUTING chains, it makes sense that we'll have to check those separately and make sure that they jump to the PROXY_INIT_OUTPUT and PROXY_INIT_REDIRECT chains.

internal/iptables/iptables_test.go
```diff
@@ -22,7 +23,99 @@ func TestMakeMultiportDestinations_Split(t *testing.T) {
...
+		{
+			name: "no existing rules, create new chain and PREROUTING rule",
+			wantCommands: []*exec.Cmd{
+				exec.Command("<iptables>", "-t", "nat", "-N", "PROXY_INIT_REDIRECT"),
+				exec.Command("<iptables>", "-t", "nat", "-A", "PROXY_INIT_REDIRECT", "-p", "tcp", "--match", "multiport", "--dports", "1234", "-j", "RETURN", "-m", "comment", "--comment", "proxy-init/ignore-port-1234/testExecutionTraceID"),
+				exec.Command("<iptables>", "-t", "nat", "-A", "PREROUTING", "-j", "PROXY_INIT_REDIRECT", "-m", "comment", "--comment", "proxy-init/install-proxy-init-prerouting/testExecutionTraceID"),
+			},
+		},
+		{
+			name:          "existing rules, flush existing chain and reuse PREROUTING rule",
```
kleimkuhler, line 61: This formatting is odd, but I see that it's automatically happening.

kleimkuhler (approved):
The tests for this look great and help understand the changes more. Thanks for going through all the back-and-forth this.

internal/iptables/iptables_test.go
jim-minter: Yes, it's automatic: gofmt is aligning with the field immediately below, hence the spaces.

jim-minter:
@stevej, please can this be merged?

stevej (approved):
Thank you!

PR #167, opened 2023-01-11
Add static and dynamic port overrides for CNI ebpf (#9841)

stevej:
When CNI plugins run in ebpf mode, they may rewrite the packet destination when doing socket-level load balancing (i.e in the `connect()` call). In these cases, skipping `443` on the outbound side for control plane components becomes redundant; the packet is re-written to target the actual Kubernetes API Server backend (which typically listens on port `6443`, but may be overridden when the cluster is created).

This change adds port `6443` to the list of skipped ports for control plane components. On the linkerd-cni plugin side, the ports are non-configurable. Whenever a pod with the control plane component label is handled by the plugin, we look-up the `kubernetes` service in the default namespace and append the port values (of both ClusterIP and backend) to the list.

On the initContainer side, we make this value configurable in Helm and provide a sensible default (`443,6443`). Users may override this value if the ports do not correspond to what they have in their cluster. In the CLI, if no override is given, we look-up the service in the same way that we do for linkerd-cni; if failures are encountered we fallback to the default list of ports from the values file.

Closes #9817

### DIFF
--- a/cni-plugin/main.go
+++ b/cni-plugin/main.go
@@ -36,9 +36,9 @@ import (
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // ProxyInit is the configuration for the proxy-init binary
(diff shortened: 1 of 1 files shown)


mateiidavid (approved):

cni-plugin/main.go
```diff
@@ -309,6 +319,23 @@ func cmdDel(args *skel.CmdArgs) error {
...
 }
 
+func getAPIServerPorts(ctx context.Context, api *kubernetes.Clientset) ([]string, error) {
+	service, err := api.CoreV1().Services("default").Get(ctx, "kubernetes", metav1.GetOptions{})
+	if err != nil {
+		return []string{}, err
+	}
+
+	ports := make([]string, 0)
+	for _, port := range service.Spec.Ports {
```
jeremychase, line 329: @stevej @mateiidavid This would be more robust if we looked at the return value of `IntValue()` rather than checking the `Type`.

For example:

```go
	for _, port := range service.Spec.Ports {
		ports = append(ports, strconv.Itoa(int(port.Port)))

		if tp := port.TargetPort.IntValue(); tp == 0 {
			logrus.Debugf("linkerd-cni: failed to convert target port, omitting: %s", port.TargetPort.String())
		} else {
			ports = append(ports, strconv.Itoa(tp))
		}
	}
```

PR #168, opened 2023-01-18
release workflow for cni-plugin.yml and updated RELEASE.md

stevej:
Github release workflow for cni-plugin that matches proxy-init. The easiest way to diff this is to checkout the branch and compare it with proxy-init: 

```
cd .github/workflows
diff -y release-proxy-init.yml release-cni-plugin.yml
```

Advice for testing this is appreciated!


mateiidavid (approved):
This looks great. 

So we could test by opening a PR with tag `cni-plugin/test` (and that would test the build basically but won't push), and then if we really want to be cautious we could manually test on a fork? afaik that won't be super straightforward since the fork will need to have a GH PAT generated, but I think on a fork the image will go to the repo owner's package registry so we could publish an image that's not shared and use it to test locally (by signing into the registry using the same PAT).

stevej:
> So we could test by opening a PR with tag `cni-plugin/test`

Interestingly, this PR results in an image: ghcr.io/linkerd/cni-plugin:test-46cff54 so I will use that for my `linkerd2` work and see what happens.

stevej:
I made a backlog item for when to run integration tests in the release workflow.

PR #176, opened 2023-01-23
updating devcontainer where dependabot missed

stevej:
TODO: figure out why dependabot missed

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -11,21 +11,21 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-20.04
-    container: ghcr.io/linkerd/dev:v38-go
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -11,21 +11,21 @@ on:
 jobs:
   lint:
```
stevej, line 12: does this need to be `go-fmt` for dependabot to inspect it?

kleimkuhler: I don't think dependabot matches based off the job name. At least, I don't' see anything in `dependabot.yml` that indicates that.

kleimkuhler (approved):

alpeb (approved):
LGTM. No clue why this wasn't picked up by dependabot :man_shrugging:

PR #177, opened 2023-01-23
use the cni test image that we've just built for our tests

stevej:
We were using an edge release before which was not testing our changes.

### DIFF
--- a/cni-plugin/main.go
+++ b/cni-plugin/main.go
@@ -130,7 +130,6 @@ func parseConfig(stdin []byte) (*PluginConf, error) {
 
 // cmdAdd is called by the CNI runtime for ADD requests
 func cmdAdd(args *skel.CmdArgs) error {
-	logrus.Debug("linkerd-cni: cmdAdd, parsing config")
 	conf, err := parseConfig(args.StdinData)
 	if err != nil {
 		logrus.Errorf("error parsing config: %e", err)
--- a/cni-plugin/main.go
+++ b/cni-plugin/main.go
@@ -82,10 +82,12 @@ type PluginConf struct {
 }
 
 func main() {
+	// Must log to Stderr because the CNI runtime uses Stdout as its state
+	logrus.SetOutput(os.Stderr)
 	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, "")
 }
 
-func configureLogging(logLevel string) {
+func configureLoggingLevel(logLevel string) {
 	switch strings.ToLower(logLevel) {
 	case "debug":
 		logrus.SetLevel(logrus.DebugLevel)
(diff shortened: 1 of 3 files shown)


cni-plugin/main.go
```diff
@@ -134,9 +133,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 	logrus.Debug("linkerd-cni: cmdAdd, parsing config")
 	conf, err := parseConfig(args.StdinData)
 	if err != nil {
+		logrus.Errorf("error parsing config: %e", err)
 		return err
 	}
-	configureLogging(conf.LogLevel)
+	configureLoggingLevel(conf.LogLevel)
```
mateiidavid, line 139: I know it's not part of this change, but I never understood why we logged on L133 and then configured the logging a few lines below it. Surely we should just configure it and then log? Would it make sense to just move this to the top of the function's scope now?

internal/iptables/iptables.go
```diff
@@ -234,7 +234,9 @@ func executeCommand(firewallConfiguration FirewallConfiguration, cmd *exec.Cmd)
 		// command.
 		//
 		// See https://github.com/rancher/k3s/issues/1434#issuecomment-629315909
-		args := append([]string{"--net", firewallConfiguration.NetNs, "--"}, cmd.Args...)
+		args := []string{fmt.Sprintf("--net=%s", firewallConfiguration.NetNs)}
+		args = append(args, "--")
+		args = append(args, cmd.Args...)
```
mateiidavid, line 239:
```go
nsArgs := fmt.Sprintf("--net=%s", firewallConfiguration.NetNs)
args := []string{nsArgs, "--", cmd.Args...)
```

An alternative way to writing this that is a bit more concise. While it probably won't do much in terms of performance (we're working with small data here) I think it's nice to have the array already initialised.

I trust your judgement more than mine so I'll leave it up to you. TIOLI :D

mateiidavid (approved):
Nice! 🚀 Almooost there. Feels great to finally be able to test all of this, and feels great to bump the proxy-init library. Can't believe we haven't been on the most recent version for what, 5 months now?

I left a few non-blocking comments for your consideration. Ship whenever.

cni-plugin/main.go
stevej: There's a chicken-and-egg situation here, you can't change the logging level until you parse the config and the debug log message on L133 exists to give us a hint that we're about to parse the config. We can remove the whole line, in my opinion.

internal/iptables/iptables.go
stevej: Is there another syntax for this?

```
internal/iptables/iptables.go:241:42: missing ',' in composite literal (typecheck)
                args := []string{nsArgs, "--", cmd.Args...}
```

mateiidavid: Oh I think I misread the code a little bit and gave a bad suggestion:

```
args := append([]string{nsArgs, "--"}, cmd.Args...)
```

this is what I had in mind

cni-plugin/main.go
mateiidavid: Yeah, let's just remove it maybe.

internal/iptables/iptables.go
stevej: Thanks! I resolved this and the question about the debug log line.

PR #178, opened 2023-01-28
Use a MultiWriter to write both to stderr and to a logfile if the pod annotation LINKERD_CNI_DEBUG_LOGFILE is passed in.

stevej:
Use a MultiWriter to write both to stderr and to a logfile if the environment variable `LINKERD_CNI_DEBUG_LOGFILE` is passed in.


cni-plugin/main.go
```diff
@@ -98,6 +99,21 @@ func configureLoggingLevel(logLevel string) {
 	}
 }
 
+// Use a MultiWriter to write both to stderr and to a logfile if
+// the environment variable LINKERD_CNI_DEBUG_LOGFILE is passed in.
+func setupDebugLog() {
+	if filename, ok := os.LookupEnv("LINKERD_CNI_DEBUG_LOGFILE"); ok {
```
stevej, line 105: I suspect this should be more aggressively named to avoid people using it in production without log rotation.

mateiidavid (approved):
Nice, this looks good to me.

mateiidavid:
I marked my previous approval as outdated. The logic looks good but when I started playing around with it I realised we don't have an easy way to actually set the environment variable.

The CNI plugin will run on the host so we can't actually enable this through the deployment. This would require us to set the environment variable on the host prior to running the plugin...which is fine, but mileage may vary. k3d for example doesn't have a .bashrc where we can persist these changes, and I suspect we'd run into issues with other distributions as well.

A somewhat more correct way might be for us to hook this value up through the installer. We already mount the host's filesystem. Perhaps we should manage the log file from the installer and then in code if the file exists we use it in the multiwriter? This might not be the best way to do it either.

I couldn't get it to work but perhaps I'm doing something different than you are? Did it work for you?

stevej:
Feedback from Matei in Slack: 
```
I'm pretty sure we could do it in a quick and clean way in the install script though:
Read the env var (turn it into a true/false value).
If true, set a new field in the conf json using jq (should be easy to inject perhaps)
Check if turned on in the cni bin
We'd follow the same flow as the log level basically.
```

cni-plugin/main.go
```diff
@@ -98,6 +99,24 @@ func configureLoggingLevel(logLevel string) {
...
+// Linkerd-cni does not manage log rotation of this file, administrators
+// who choose to enable this for production use SHOULD provide log rotation.
+//
+// Use a MultiWriter to write both to stderr and to a logfile if
+// the environment variable LINKERD_CNI_DEBUG_LOGFILE is passed in.
+func setupDebugLog(pod *v1.Pod) {
+	if filename := pod.GetObjectMeta().GetAnnotations()["LINKERD_CNI_DEBUG_LOGFILE"]; filename != "" {
+		f, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE, 0644) //nolint:gosec
+		if err != nil {
+			panic(fmt.Sprintf("failed to create debug log file at %s with err %e", filename, err))
```
mateiidavid, line 113: If we end up doing it this way we might want to rethink this panic. It made sense when we were setting everything up, but I don't think we should panic in the middle of the pod processing.

For example, if in a more restricted environment we cannot create a debug fail, we would end up failing the whole operation -- side effect probably outweighs the good this function would do.

cni-plugin/main.go
```diff
@@ -188,6 +207,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 			logrus.Errorf("linkerd-cni client err in client.Pods().Get(): %e", err)
 			return err
 		}
+		setupDebugLog(pod)
```
mateiidavid, line 210: I think it's still helpful to log stuff to a file past this point; most of the failure points occur from here on out. How I wish it'd be simpler to do it earlier though... feels like we can mix some valuable context such as configuration options, errors received from improper kubeconfig, API Server errors when fetching the pod... I'm fine personally with what we'd be missing out on, since we at least have more visibility than before, and it makes it easier to debug, but perhaps this should only be an interim fix until we find a more robust configuration method.

cni-plugin/integration/run.sh
```diff
@@ -99,6 +99,11 @@ echo 'PASS: Network Validator'
 
 calico_overrides="{
                \"apiVersion\": \"v1\",
+               \"metadata\": {
+                \"annotations\": {
+                    \"LINKERD_CNI_DEBUG_LOGFILE\": \"/var/log/linkerd-cni.log\",
```
mateiidavid, line 104: we'd probably want to change the format here a bit. k8s annotation follow a more strict domain convention (think JVM packages)

```
config.linkerd.io/cni-debug-logfile: "/var/log/linkerd-cni.log"
```

Could also introduce a new subdomain? `debug.linkerd.io/cni-logfile`.

stevej:
After discussing with matei, we're shelving this as we have enough debugging facilities thanks to our recent work.

PR #182, opened 2023-02-01
Add Calico scenario

mateiidavid:
First take at adding Calico tests. We modify the test harness to check for the "calico" scenario; when turned on, we will apply the manifest. For the first pass, we made the following changes:

1. Harness now will check the value of SCENARIO, and will apply the manifest as previously substantiated.
2. A new go test with calico specific config is added. This does not differ from flannel except for the different dir and filename
3. Pipeline is split into three: proxy-init, cni-flannel and cni-calico. All jobs should run in parallel. We build the CNI plugin images twice. I don't anticipate this being an issue since the images are small. We could use `upload-artifact` and `download-artifact` to mitigate this. When we introduce Cilium, I suspect this will be a better way to handle the build (i.e introduce a prerequisite that builds & saves images as tars, then in parallel, run each scenario & load the tar into docker).


cni-plugin/integration/calico/calico_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if !*runTests {
+		fmt.Fprintln(os.Stderr, "integration tests not enabled: enable with -integration-tests")
+		os.Exit(0)
+	}
+
+	os.Exit(m.Run())
+}
+
+func TestLinkerdIsLastCNIPlugin(t *testing.T) {
```
stevej, line 43: This plus the associated functions would be great to pull into a test helper

cni-plugin/integration/run.sh
```diff
@@ -29,6 +29,12 @@ function cleanup() {
     k delete ns cni-plugin-test || echo "could not delete namespace cni-plugin-test"
 }
 
+function install_calico() {
+  echo '# Installing Calico...'
+  local yaml="https://k3d.io/v5.3.0/usage/advanced/calico.yaml"
+  k apply -f $yaml || echo "could not apply $yaml"
```
stevej, line 35: for cleanup, you can add a function `cleanup_calico` and then call `trap cleanup_calico EXIT` here

cni-plugin/integration/run.sh
```diff
@@ -59,16 +71,11 @@ k run linkerd-proxy \
     --restart=Never \
     --rm \
     -- \
-    /usr/lib/linkerd/linkerd2-network-validator --log-format plain \
```
stevej, line 62: this is probably an accidental deletion that is causing CI to fail.

.github/workflows/integration.yml
```diff
@@ -11,17 +11,33 @@ on:
       - proxy-init/**
 
 jobs:
-  integration:
+  proxy-init-integration:
     timeout-minutes: 15
```
stevej, line 15: would you mind adding `continue-on-error: true` here as well, I think that's a good fix.

stevej (changes requested):
This is looking great so far!

cni-plugin/integration/calico/calico_test.go
```diff
@@ -0,0 +1,90 @@
...
+	t.Parallel()
```
alpeb, line 44: I believe this isn't needed given both tests (flannel and calico) are triggered through separate `go test` invocations.

alpeb (commented):
Looking good! Did you miss changes to the `justfile` accounting for the `SCENARIO` env var?

cni-plugin/integration/calico/calico_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+		var conflist map[string]any
+		err = json.Unmarshal(conflistFile, &conflist)
+		if err != nil {
+			t.Fatalf("unmarshaling conflist json failed: %e", err)
+		}
+
+		if conflist["cniVersion"] != "1.0.0" {
+			t.Fatalf("expected cniVersion 1.0.0, instead saw %s", conflistFile)
+		}
```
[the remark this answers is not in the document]
stevej: Yeah, I think that's true; this is copied from the flannel unit tests I wrote which matei is going to abstract into a test helper.

cni-plugin/integration/calico/calico_test.go
stevej: That's true and when we write more tests we can think about adding the line back.

cni-plugin/integration/run.sh
mateiidavid: Nice, I ended up doing it all in the same cleanup, perhaps a separate one might make more sense though... lmk what you think.

cni-plugin/integration/run.sh
mateiidavid: Re-added.

cni-plugin/integration/run.sh
```diff
@@ -60,31 +94,46 @@ k run linkerd-proxy \
     --rm \
     -- \
     /usr/lib/linkerd/linkerd2-network-validator --log-format plain \
-      --log-level debug --connect-addr 1.1.1.1:20001 \
-      --listen-addr 0.0.0.0:4140 --timeout 10s
+    --log-level debug --connect-addr 1.1.1.1:20001 \
+    --listen-addr 0.0.0.0:4140 --timeout 5m
```
stevej, line 98: Were you seeing issues that required raising this timeout? I tried to stick with what we're deploying in customer's clusters.

cni-plugin/integration/testutil/test_util.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("could not read %s: %w", r.confFile, err)
+	}
+
+	var conflist map[string]any
+	err = json.Unmarshal(conflistFile, &conflist)
+	if err != nil {
+		return fmt.Errorf("unmarshaling conflist json failed: %w", err)
+	}
+
+	// if conflist["cniVersion"] != "1.0.0" {
```
stevej, line 79: If we don't rely on CNI version 1.0.0, we should probably just delete this. Is there a minimum CNI version we expect?

cni-plugin/integration/tests/calico/calico_test.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/linkerd/linkerd2-proxy-init/cni-plugin/integration/testutil"
+)
+
+const (
+	ConfigDirectory = "/host/etc/cni/net.d"
+	CalicoConflist  = "10-calico.conflist"
```
stevej, line 12: Calico ships a `.conflist` and not a `.conf` That's great, that removes one of the Docker-based tests then.

cni-plugin/integration/run.sh
```diff
@@ -19,14 +23,41 @@ function create_test_lab() {
     k create serviceaccount linkerd-cni
     # TODO(stevej): how can we parameterize this manifest with `version` so we
     # can enable a testing matrix?
-    k create -f manifests/linkerd-cni.yaml
+    # Apply all files in scenario directory. For non-flannel CNIs, this will
+    # include the CNI manifest itself.
+    k apply -f "manifests/$SCENARIO/"
 }
 
+# TODO (matei): skip this based on env var? Useful when running locally to see
```
stevej, line 31: What I do when developing is add a `sleep 60` and then catch it with `Ctrl-Z` to suspend `run.sh` That allows me to inspect all the logs and `describe` everything. Once I've collected enough data, I use `fg` and then cleanup occurs. I'm sure we can think of a cleaner way to do this but probably in a later PR. maybe `kill -TSTP %1`?

cni-plugin/integration/run.sh
mateiidavid: You know what, you were completely right. I thought this timeout was passed to `kubectl` not the validator 🤦🏻

cni-plugin/integration/tests/calico/calico_test.go
mateiidavid: It seems to ship with a `.conflist`, at least the version we're using here.

```
:; docker exec -it k3d-calico-server-0 sh
/ # cd /etc/
/etc # ls
cni  hostname  hosts  mtab  nsswitch.conf  os-release  rancher  resolv.conf  ssl  strongswan
/etc # cd cni/net.d/
/etc/cni/net.d # ls
10-calico.conflist  calico-kubeconfig
/etc/cni/net.d #
```

cni-plugin/integration/testutil/test_util.go
stevej: Can we add a test-only dependency on https://github.com/coreos/go-semver ?

justfile
```diff
@@ -145,24 +145,55 @@ build-cni-plugin-test-image *args='--load':
...
-# Run cni-plugin integration tests after preparing dependencies
-# For new scenarios, add them after cni-plugin-test-integration-deps
-cni-plugin-test-integration: cni-plugin-test-integration-deps cni-plugin-test-integration-flannel
 
-# Run flannel integration tests without preparing dependencies
-cni-plugin-test-integration-flannel:
-    SCENARIO=flannel TEST_CTX="k3d-$(just-k3d --evaluate K3D_CLUSTER_NAME)" ./cni-plugin/integration/run.sh
+# Run cni-plugin integration tests after preparing dependencies By default,
+# runs "flannel" scenario, behavior can be overridden through
+# `cni-integration-scenario` variable
```
alpeb, line 155: I couldn't find where this variable is used.

cni-plugin/integration/manifests/calico/calico-install.yaml
```diff
@@ -0,0 +1,3755 @@
+---
+# Source: calico/templates/calico-config.yaml
```
alpeb, line 2: TIOLI: Are all these manifests taken from a single URL, or is this file the result of concatenating multiple files? If the latter, it might be better to put them separately under a directory, mirroring the layout upstream, to more easily upgrade when the time comes...

alpeb (commented):
Sweet!

cni-plugin/integration/manifests/calico/calico-install.yaml
stevej: I did this part so I can explain: they are the output of `linkerd install-cni` with appropriate arguments.

justfile
```diff
@@ -145,24 +145,55 @@ build-cni-plugin-test-image *args='--load':
...
+# To run all scenarios concurrently, see: `cni-plugin-test-integration-all`
+cni-plugin-test-integration: _cni-plugin-test-integration-deps _cni-plugin-test-integration
 
+# Run all integration test scenarios, in different environments
+cni-plugin-test-integration-all: cni-plugin-test-integration-flannel cni-plugin-test-integration-calico
```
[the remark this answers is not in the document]
mateiidavid: Ah yeah, you're right. Don't think you can make them run in parallel actually, or at least I haven't been able to figure it out from the man.

justfile
mateiidavid: Yeah I removed it and forgot to update the docs. It was replaced by an env var.

cni-plugin/integration/manifests/calico/calico-install.yaml
mateiidavid: It's the former. Originally we applied it straight from the internet...but! Lossy networks, unreliable links, and complicated logic made me think it's better to just `curl` a copy locally. It makes it easier to maintain a directory for each scenario and apply it _all_ instead of maintaining separate logic branches that are based on the scenario being run.

I think in the long run your concern is well founded though, we'll need to find a way to upgrade or version calico. We should perhaps move to `helm` at some point. If we ever want to do more advanced testing (e.g ebpf mode, different other modes such as policy engines) we'll need a more expressive way of installing anyway.

alpeb: Cool, this should do for now then :+1:

alpeb (approved):
This looks good to me! :taco:

cni-plugin/integration/run.sh
stevej: I'll put a comment in here while I'm making the next test. Nothing to do on your part.

stevej (approved):
LGTM!

PR #189, opened 2023-02-14
Check that the linkerd config embedded in cni config is correct

stevej:

### DIFF
--- a/cni-plugin/integration/Dockerfile-tester
+++ b/cni-plugin/integration/Dockerfile-tester
@@ -9,6 +9,7 @@
 #  2) a test suite (e.g. `flannel.go`) runs with a configured CNI plugin.
 
 FROM golang:1.18-alpine AS build
+RUN apk add build-base
 ENV GOCACHE=/tmp/
 WORKDIR /src
 COPY --link go.mod go.sum .
--- a/cni-plugin/integration/testutil/test_util.go
+++ b/cni-plugin/integration/testutil/test_util.go
@@ -144,7 +144,7 @@ func (r *TestRunner) CheckCNIPluginIsLast() error {
 	plugins := conflist["plugins"].([]interface{})
 	lastPlugin := plugins[len(plugins)-1].(map[string]any)
 	if err = checkLinkerdCniConf(lastPlugin); err != nil {
-		return fmt.Errorf("Configuration contains erroneous value\n%s", err)
+		return fmt.Errorf("Configuration contains erroneous value\n%w", err)
 	}
 	if lastPlugin["name"] != "linkerd-cni" {
 		return fmt.Errorf("linkerd-cni was not last in the plugins list")
--- a/cni-plugin/integration/testutil/test_util.go
+++ b/cni-plugin/integration/testutil/test_util.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strconv"
 
 	"github.com/mitchellh/mapstructure"
 )
@@ -54,6 +53,16 @@ type ProxyInit struct {
 	UseWaitFlag           bool     `json:"use-wait-flag"`
 }
 
+// LinkerdPlugin is what we use for CNI configuration in the plugins section
+type LinkerdPlugin struct {
+	Name       string         `json:"name"`
+	Type       string         `json:"type"`
+	LogLevel   string         `json:"log_level"`
+	Policy     map[string]any `json:"policy"`
+	Kubernetes map[string]any `json:"kubernetes"`
+	ProxyInit  ProxyInit      `json:"linkerd"`
+}
+
 // Checks that the embedded linkerd config is of the expected form
 // and contains the right values:
 //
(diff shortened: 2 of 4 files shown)


cni-plugin/integration/testutil/test_util.go
```diff
@@ -39,6 +40,70 @@ func (r *TestRunner) walkConfDir() (map[string]struct{}, error) {
...
+//	   "proxy-uid": 2102,
+//	   "ports-to-redirect": [],
+//	   "inbound-ports-to-ignore": ["4191","4190"],
+//	   "simulate": false,
+//	   "use-wait-flag": false
+//	 }
+func checkLinkerdCniConf(wrapperConf map[string]any) error {
+	var conf = wrapperConf["linkerd"].(map[string]any)
+
+	var incomingProxyPort = conf["incoming-proxy-port"].(float64)
```
mateiidavid, line 58: Genuine question: is there a difference to using a float64 for the port as opposed to using an int?

cni-plugin/integration/testutil/test_util.go
```diff
@@ -39,6 +40,70 @@ func (r *TestRunner) walkConfDir() (map[string]struct{}, error) {
...
+	if incomingProxyPort != 4143 {
+		return fmt.Errorf("incoming-proxy-port has wrong value: expected: %v, found: %v",
```
mateiidavid, line 60: Maybe the linter will complain, but we could use %d as the format directive

cni-plugin/integration/testutil/test_util.go
```diff
@@ -78,6 +143,10 @@ func (r *TestRunner) CheckCNIPluginIsLast() error {
 
 	plugins := conflist["plugins"].([]interface{})
 	lastPlugin := plugins[len(plugins)-1].(map[string]any)
+	err = checkLinkerdCniConf(lastPlugin)
+	if err != nil {
+		return err
+	}
```
mateiidavid, line 149: TIOLI:

```suggestion
	if err = checkLinkerdCniConf(lastPlugin); err != nil {
		return fmt.Errorf("Configuration contains erroneous value\n%s", err)
	}
```

I inlined the if statement (tbh I'm not the biggest fan of this but it makes sense here perhaps?) and then changed the error message to reflect where in the test the failure happened. By no means an improvement, just a suggestion, feel free to ignore both or either of them :)

cni-plugin/integration/testutil/test_util.go
mateiidavid, line 60:
```suggestion
		return fmt.Errorf("incoming-proxy-port has wrong value, expected: %v, found: %v",
```

Just a tiny nitpick, noticed the rest use a comma after.

cni-plugin/integration/testutil/test_util.go
stevej: I think this is because JSON only has floating point numbers. When I try to use `int` I get the following error

`panic: interface conversion: interface {} is float64, not int [recovered]`

cni-plugin/integration/testutil/test_util.go
stevej: with an `int` it will work but not with a `float64`

cni-plugin/integration/testutil/test_util.go
mateiidavid: eesh :(

mateiidavid (approved):
Nice!

cni-plugin/integration/testutil/test_util.go
```diff
@@ -39,6 +40,70 @@ func (r *TestRunner) walkConfDir() (map[string]struct{}, error) {
...
+//	"linkerd": {
+//	   "incoming-proxy-port": 4143,
+//	   "outgoing-proxy-port": 4140,
+//	   "proxy-uid": 2102,
+//	   "ports-to-redirect": [],
+//	   "inbound-ports-to-ignore": ["4191","4190"],
+//	   "simulate": false,
+//	   "use-wait-flag": false
+//	 }
+func checkLinkerdCniConf(wrapperConf map[string]any) error {
```
mateiidavid, line 55: Not a blocker but should we create a wrapper struct to more easily deserialize this config in a follow-up PR? Do you think that'd make things a bit easier to work with? Something similar to https://github.com/linkerd/linkerd2-proxy-init/blob/main/cni-plugin/main.go#L44-L55. Perhaps we could even move that in a shared (internal) library 🤷🏻 

This test is good to assert default values are fine, but we might want to try with some overrides in the future?

mateiidavid (approved):

cni-plugin/integration/testutil/test_util.go
stevej: I think using the struct makes sense but I can't figure out how to import `ProxyInit` from `main.go` into this test package.

stevej: I switched to using `mapstructure` which has more options for dynamically decoding structs, take a look.

cni-plugin/integration/Dockerfile-tester
```diff
@@ -9,6 +9,7 @@
 #  2) a test suite (e.g. `flannel.go`) runs with a configured CNI plugin.
 
 FROM golang:1.18-alpine AS build
+RUN apk add build-base
```
mateiidavid, line 12: Were we missing tooling from alpine-go base image?

cni-plugin/integration/testutil/test_util.go
```diff
@@ -77,10 +175,14 @@ func (r *TestRunner) CheckCNIPluginIsLast() error {
 	}
 
 	plugins := conflist["plugins"].([]interface{})
-	lastPlugin := plugins[len(plugins)-1].(map[string]any)
-	if lastPlugin["name"] != "linkerd-cni" {
+	linkerdPlugin := plugins[len(plugins)-1].(map[string]any)
```
mateiidavid, line 178: Do you think we could use a comment here? I wanted to suggest it but tbh it's pretty self-explanatory if we look at the error message on L180.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/containernetworking/cni v1.1.2
+	github.com/mitchellh/mapstructure v1.1.2
```
mateiidavid, line 7: wish Go had dev deps :(

mateiidavid (approved):

cni-plugin/integration/Dockerfile-tester
stevej: When I added `mapstructure`, the container failed with `Cgo` errors which I tracked down to needing something provided by `build-base`, gcc I think.

cni-plugin/integration/testutil/test_util.go
stevej: comment added!

mateiidavid:
LGTM, will merge this.

PR #194, opened 2023-03-01
Introduce a running mode for CNI plugin

mateiidavid:
Currently, the CNI plugin will adapt to its environment dynamically, depending on the presence of a configuration file. For example, when a configuration file is present, it simply appends itself to it. Otherwise, it will create its own configuration file.

In certain scenarios, such as when a conflist file is deleted, inotify mgiht not be quick enough to capture the change, and as such will attempt to install linkerd-cni in interface mode. After this, it will attempt to delete its own configuration file, but since it has already been observed by inotify, it will compute the diff hash and think it has to be re-created, thereby getting into an endless loop.

An even better solution to this would be to completely delete standalone mode, since that does not add much to the cni plugin, as noted in #10375. Linkerd CNI would probably almost never run in standalone mode. For backwards compatibility, it is still possible; by default, plugin runs in chained mode.

fixes #linkerd2/10375

### DIFF
--- a/cni-plugin/deployment/scripts/install-cni.sh
+++ b/cni-plugin/deployment/scripts/install-cni.sh
@@ -56,6 +56,9 @@ HOST_CNI_NET="${CONTAINER_MOUNT_PREFIX}${DEST_CNI_NET_DIR}"
 # Default path for when linkerd runs as a standalone CNI plugin
 DEFAULT_CNI_CONF_PATH="${HOST_CNI_NET}/01-linkerd-cni.conf"
 KUBECONFIG_FILE_NAME=${KUBECONFIG_FILE_NAME:-ZZZ-linkerd-cni-kubeconfig}
+# Mode to install plugin in
+# Unless overridden, we install linkerd-cni in 'chain' mode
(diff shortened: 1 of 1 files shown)


cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -323,23 +326,32 @@ monitor() {
...
-else
-  find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
-    while read -r -d $'\0' file; do
-      echo "Installing CNI configuration in \"chained\" mode for $file"
-      install_cni_conf "$file"
-    done
+  while true; do
+    sleep infinity &
+    wait $!
+  done
```
alpeb, line 337: Why does this require to be wrapped in a loop?

alpeb:
While testing I realized that (in the default "chained mode") deleting the cni config file triggers `sync()` and recreates the file in "interface" mode. I guess that logic needs to be updated so it waits for the original file again? Not a likely scenario, but for consistency sake :man_shrugging:

mateiidavid:
In draft while we figure out a testing plan.

PR #200, opened 2023-03-08
tests cilium interop with linkerd-cni

stevej:
We add cilium to the mix!

Adds the following: 
* justfile targets for running a cilium installer and special fs setup
* test runner for checking that the config file is ok
* workflow integration for CI

### DIFF
--- a/cni-plugin/integration/run.sh
+++ b/cni-plugin/integration/run.sh
@@ -96,7 +96,7 @@ k run linkerd-proxy \
 
 echo 'PASS: Network Validator'
 
-calico_overrides="{
+generic_config_mount="{
                \"apiVersion\": \"v1\",
                \"spec\": {
                   \"containers\": [
@@ -123,7 +123,7 @@ calico_overrides="{
                },
                \"status\": {}
             }"
-flannel_overrides="{
+flannel_config_mount="{
                \"apiVersion\": \"v1\",
                \"spec\": {
                   \"containers\": [
@@ -151,7 +151,6 @@ flannel_overrides="{
                \"status\": {}
             }"
 # This needs to use the name linkerd-proxy so that linkerd-cni will run.
-# We use the calico overrides for cilium to avoid duplication.
 echo '# Running tester...'
 if [ "$SCENARIO" == "calico" ] || [ "$SCENARIO" == "cilium" ]; then
   k run linkerd-proxy \
@@ -160,7 +159,7 @@ if [ "$SCENARIO" == "calico" ] || [ "$SCENARIO" == "cilium" ]; then
           --image-pull-policy=Never \
           --namespace=cni-plugin-test \
           --restart=Never \
-          --overrides="$calico_overrides" \
+          --overrides="$generic_config_mount" \
           --rm
 else
   k run linkerd-proxy \
@@ -169,6 +168,6 @@ else
           --image-pull-policy=Never \
           --namespace=cni-plugin-test \
           --restart=Never \
-          --overrides="$flannel_overrides" \
+          --overrides="$flannel_config_mount" \
           --rm
 fi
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -178,6 +181,35 @@ cni-plugin-test-integration-calico:
...
+        cni-plugin-test-integration-with-cilium
+
+_cni-plugin-setup-cilium:
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    docker exec k3d-l5d-cilium-test-server-0 mount bpffs /sys/fs/bpf -t bpf
+    docker exec k3d-l5d-cilium-test-server-0 mount --make-shared /sys/fs/bpf
+    docker exec k3d-l5d-cilium-test-server-0 mkdir -p /run/cilium/cgroupv2
+    docker exec k3d-l5d-cilium-test-server-0 mount -t cgroup2 none /run/cilium/cgroupv2
+    docker exec k3d-l5d-cilium-test-server-0 mount --make-shared /run/cilium/cgroupv2/
```
mateiidavid, line 198: could probably do all of these in one go:

```
docker exec k3d-l5d-cilium-test-server-0 "mount bpffs /sys/fs/bpf -t bpf; \
 mount --make-shared /sys/fs/bpf;\
 ...
"
```

justfile
```diff
@@ -178,6 +181,35 @@ cni-plugin-test-integration-calico:
...
+    echo "Mounted /sys/fs/bpf to cilium-test-server cluster"
+    helm repo add cilium https://helm.cilium.io/
+    helm install cilium cilium/cilium --version 1.13.0 \
```
mateiidavid, line 201: We could maybe silence everything except the echos with @? wdyt?

cni-plugin/integration/run.sh
```diff
@@ -152,8 +151,9 @@ flannel_overrides="{
                \"status\": {}
             }"
 # This needs to use the name linkerd-proxy so that linkerd-cni will run.
+# We use the calico overrides for cilium to avoid duplication.
 echo '# Running tester...'
-if [ "$SCENARIO" == "calico" ]; then
+if [ "$SCENARIO" == "calico" ] || [ "$SCENARIO" == "cilium" ]; then
```
jeremychase, line 156: thought: Does `--overrides="$calico_overrides" \` have implications here? Just wondering if this should be another else if block rather than grouped with `calico`.

justfile
stevej: I think these need to be on separate lines in order for the operations to occur in-order. I don't think mount is seeing the result if mkdir and here's the error I get when I combine them:

```+ docker exec k3d-l5d-cilium-test-server-0 'mount bpffs /sys/fs/bpf -t bpf;    mount --make-shared /sys/fs/bpf;    mkdir -p /run/cilium/cgroupv2;    mount -t cgroup2 none /run/cilium/cgroupv2;    mount --make-shared /run/cilium/cgroupv2/'
OCI runtime exec failed: exec failed: unable to start container process: exec: "mount bpffs /sys/fs/bpf -t bpf;    mount --make-shared /sys/fs/bpf;    mkdir -p /run/cilium/cgroupv2;    mount -t cgroup2 none /run/cilium/cgroupv2;    mount --make-shared /run/cilium/cgroupv2/": stat mount bpffs /sys/fs/bpf -t bpf;    mount --make-shared /sys/fs/bpf;    mkdir -p /run/cilium/cgroupv2;    mount -t cgroup2 none /run/cilium/cgroupv2;    mount --make-shared /run/cilium/cgroupv2/: no such file or directory: unknown

cni-plugin/integration/run.sh
stevej: I could break it into two if blocks but `cilium_overrides` would be 100% identical to `calico_overrides` and that bothered me.

I think what makes sense here is to rename `calico_overrides` to `generic_config_mount` and I've gone ahead and done that.

PR #204, opened 2023-03-13
delete the old docker tests, github workflow, and justfile target

stevej:
With the new flannel, calico, and cilium tests, we no longer need the old docker-based tests.


alpeb (approved):
Good riddance! :+1: :100: :1st_place_medal:

PR #237, opened 2023-05-02
cni: use `scratch` as the base runtime docker image

hawkw:
Currently, the CNI plugin Docker image uses a Debian base image at runtime. Debian is used to install some packages that are dependencies of the `install-cni.sh` script, namely `inotifywatch`, `jq`, and `pgrep`. However, these packages are the only things we need, and it's not strictly necessary to run the CNI plugin in an entire Debian install. We could install these tools from a Debian image, and then actually run in a `scratch` image.

This branch changes the CNI plugin `Dockerfile` to use a `scratch` base image for the final runtime layer. `debian:bullseye-slim` is still used when building the image, in order to install the required packages using APT, and then the installed binaries are copied into the final image.

This is the same change as linkerd/linkerd2#10845, but applied to the dockerfile in this repo instead.


alpeb (approved):
LGTM :+1: :crossed_fingers:

PR #238, opened 2023-05-04
cni: change base runtime image from `scratch` to `alpine`

hawkw:
PR #237 changed the base runtime image for the CNI init container from `debian:bullseye-slim` to `scratch`, in the hopes of reducing the dependency footprint of that image. Unfortunately, it turns out that this breaks the CNI init container, since it must execute shell scripts and therefore requires a shell (specifically, `bash`), which does not exist in the `scratch` image. This was not detected by CI in this repo on PR #237, because it turns out that the CI jobs don't actually try to run the CNI init container. Instead, we didn't discover that this breaks stuff until the change was fully integrated in linkerd/linkerd2#10855.

This commit changes the base runtime image again, to `alpine:3.17.3`, similarly to the proxy-init dockerfile. We now install the necessary dependencies using `apk add` in the alpine layer. This way, we should have a shell, and be able to run the install script again.

### DIFF
--- a/Dockerfile-cni-plugin
+++ b/Dockerfile-cni-plugin
@@ -20,7 +20,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH GO111MODULE=on \
 ## Runtime
 ##
 
-FROM --platform=$TARGETPLATFORM alpine:3.17.3 as runtime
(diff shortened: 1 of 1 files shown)


hawkw:
huh, that's weird, it claims `procps-ng` doesn't exist? https://pkgs.alpinelinux.org/package/edge/main/armhf/procps-ng

PR #304, opened 2023-12-01
Update build tooling

alpeb:
- Updated rust toolchain to 1.73.0
- Updated `dev` container to v42, which also implied:
  - Removed the `depguard` go lint
  - Removed dupe --no-run flag in justfile
- Extracted duped `meta` job into separate composite action
- Added `deny.toml`
- Updated kubert and clap dependencies
- Moved `./validator/.justfile` to `./justfile-rust` and made it more generic by introducing the `TARGETCRATE` env var

This serves as a preliminary to the upcomging
`linkerd-reinitialize-pods` controller that'll live in this repo as well.


.github/workflows/go.yml
```diff
@@ -8,25 +8,31 @@ on:
...
+      - run: git config --global --add safe.directory "$PWD" # actions/runner#2033
       - run: just go-fmt-check
 
   unit-test:
     runs-on: ubuntu-latest
-    container: ghcr.io/linkerd/dev:v39-go
+    container: ghcr.io/linkerd/dev:v42-go
     steps:
       - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
+      - run: git config --global --add safe.directory "$PWD" # actions/runner#2033
```
mateiidavid, line 36: probably a bit out of scope but why do we need this now? pure curiosity from my side.

justfile-rust
```diff
@@ -1,12 +1,15 @@
...
+# This justfile includes recipes for building and packaging the validator and
+# reinitialize-pods crates for release. This file is separated so that we can
+# invoke cargo, etc when building defaults. If this logic were in the main
+# justfile, then these tools would be invoked (possibly updating Rust, etc),
+# for unrelated targets.
 #
-# Users are expected to interact with this via the top-level Justfile.
+# Users are expected to interact with this via the main Justfile.
+
+crate := env_var('TARGETCRATE')
```
mateiidavid, line 9: should we have a default here?

mateiidavid (approved):
LGTM 👍🏻

.github/workflows/go.yml
alpeb: Without this, the lint fails when the code was checked out with a recent git version. Something related to permissions:
https://github.com/actions/runner/issues/2033
Failure example:
https://github.com/linkerd/linkerd2-proxy-init/actions/runs/7061243421/job/19223102061

justfile-rust
alpeb: I'd like the env var to be required, so this would fail with a clear error if the var is not explicitly set.

PR #306, opened 2023-12-05
cni-repair controller

alpeb:
Fixes linkerd/linkerd2#11073

This fixes the issue of injected pods that cannot acquire proper network config because `linkerd-cni` and/or the cluster's network CNI haven't fully started. They are left in a permanent crash loop and once CNI is ready, they need to be restarted externally, which is what this controller does.

This controller "`linkerd-cni-repair-controller`" watches over events on pods in the current node, which have been injected but are in a terminated state and whose `linkerd-network-validator` container exited with code 95, and proceeds to delete them so they can restart with a proper network config.

The controller is to be deployed as an additional container in the `linkerd-cni` DaemonSet (addressed in linkerd/linkerd2#11699).

This exposes two custom counter metrics: `linkerd_cni_repair_controller_queue_overflow` (in the spirit of the destination controller's `endpoint_updates_queue_overflow`) and `linkerd_cni_repair_controller_deleted`


validator/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = false
 anyhow = "1"
 bytes = "1"
 kubert = { version = "0.19.0", features = ["log"] }
+linkerd-reinitialize-pods = { path = "../reinitialize-pods" }
```
olix0r, line 13: This means that the validator will now take on all of the dependencies of the controller, which isn't desirable if we're doing this to reuse a constant. It would be better to leave the validator unchanged.

reinitialize-pods/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[dependencies.clap]
+version = "4"
+default-features = false
+features = ["derive", "env", "help", "std"]
+
+[dependencies.kubert]
+version = "0.19.0"
+default-features = false
+features = ["clap", "metrics", "runtime"]
```
olix0r, line 32: You need the "admin" feature as well to get the admin server.

reinitialize-pods/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+openssl-tls = ["kube/openssl-tls"]
+# Vendor openssl to statically link lib
+openssl-vendored = ["openssl/vendored"]
+rustls-tls = ["kube/rustls-tls"]
+
+[dependencies]
+anyhow = "1"
+futures-util = "0.3"
+k8s-openapi = { version = "0.19", features = ["v1_21"] }
+kube = { version = "0.85", features = ["runtime"] }
```
olix0r, line 20: there are updated versions of k8s-openapi, kube, and kubert available (0.20, 0.87.1, and 0.20)

reinitialize-pods/Cargo.toml
alpeb: It's weird, but I'm still a 404 when hitting `/metrics`. I also tried adding both `admin` and `server` and still nothing. Note though that the probes are being served fine. Any ideas what could be going on?

reinitialize-pods/Cargo.toml
alpeb: I was hesitant to bump k8s-openapi to 0.20 because its earliest supported k8s is `v1_22` and we still officially support 1.21, but I guess it should be fine, I think we're not using any API here that would break in between those two versions.

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+                let maybe_already_evicting =
+                    pod.status.clone().and_then(|x| x.conditions).and_then(|x| {
+                        x.into_iter().find(|y| {
+                            y.reason
+                                .as_ref()
+                                .is_some_and(|z| z == CONDITION_EVICTED_REASON)
+                        })
+                    });
+
```
mateiidavid, line 42: I think we could potentially simplify this if we return early at the start when the status is missing? We don't expect that to happen often anyway, but as a precaution we could continue iterating over events:

```rust
if let watcher::Event::Applied(pod) = evt {
                let status = if let Some(ref status) = pod.status {
                    status.clone()
                } else {
                    tracing::info!("Skipped, no status");
                    continue;
                };
```

I have also tried to restructure the `maybe_terminated` and `maybe_already_evicting` lookups to avoid clones and to return a bool:

```rust
             let terminated = status
                    .init_container_statuses
                    .unwrap_or_default()
                    .iter()
                    .find(|x| x.name == "linkerd-network-validator")
                    .and_then(|x| x.last_state.as_ref())
                    .and_then(|x| x.terminated.as_ref())
                    .map(|term| term.exit_code == UNSUCCESSFUL_EXIT_CODE)
                    .unwrap_or(false);

                let evicted = status.conditions.as_ref().and_then(|xs| {
                    xs.iter().find(|y| {
                        y.reason
                            .as_ref()
                            .is_some_and(|z| z == CONDITION_EVICTED_REASON)
                    })
                });
```

We could probably use `find()` in terminated to operate on an option instead of using the filter adapter. We know we will have at most one validator. The rest is similar to your solution. I'm not adamant on my solution, I think yours reads well too  but leaving here in case it's helpful :)

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+                    let pods = kube::Api::<Pod>::namespaced(
+                        client.clone(),
+                        &pod.metadata.namespace.unwrap(),
+                    );
+                    let evict_res = pods
+                        .evict(&pod.metadata.name.clone().unwrap(), &Default::default())
+                        .await;
+                    match evict_res {
+                        Ok(_) => tracing::info!("Evicting pod {}", pod.metadata.name.unwrap()),
+                        Err(err) => tracing::warn!("Error evicting pod: {:?}", err),
```
mateiidavid, line 53: We can return the error as a field in the log. We should maybe include some sort of a key? e.g. a namespace/name to know where to look in case of errors.

```suggestion
                        Err(error) => tracing::warn!(%error, "Error evicting pod"),
```

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+                if maybe_terminated.is_some() && maybe_already_evicting.is_none() {
+                    let pods = kube::Api::<Pod>::namespaced(
+                        client.clone(),
+                        &pod.metadata.namespace.unwrap(),
+                    );
+                    let evict_res = pods
+                        .evict(&pod.metadata.name.clone().unwrap(), &Default::default())
+                        .await;
+                    match evict_res {
+                        Ok(_) => tracing::info!("Evicting pod {}", pod.metadata.name.unwrap()),
```
mateiidavid, line 52: Should this be at an info level, or should we move it to debug? Do we think this will result in a lot of spam? Also, I'd suggest using structured fields in logs maybe to make them easier to export in other formats (and to follow what we do in other controllers mostly lol).

reinitialize-pods/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
mateiidavid, line 1: hmm do we need this? won't this compile in the workspace's target dir? e.g. the proxy init root.

mateiidavid (commented):
This looks good! Exciting to finally have it committed. Two final questions:

1. Do we anticipate needing to emit events? I'm thinking for successful evictions.
2. Do we need to retry evictions? Would it ever fail due to a transient request and we wouldn't process it again? 

Both of those are more follow-ups if they're required at all. I think what we have now is great 👍🏻

reinitialize-pods/.gitignore
alpeb: Good catch, just a left-over from when I had this in a separate repo on its own :-)

reinitialize-pods/src/lib.rs
alpeb: Nice suggestions, that reads better indeed :+1:

reinitialize-pods/src/lib.rs
alpeb: Will do :+1:

reinitialize-pods/src/lib.rs
alpeb: Good point about fields :+1: 
As for the level, this entry informs about the main goal of the controller (evicting), so I'd be inclined to leave it at `info`...

alpeb:
> 1. Do we anticipate needing to emit events? I'm thinking for successful evictions.

I was expecting k8s itself to surface eviction events, but after checking it turns out that's not the case, so I'll be adding that.

> 2. Do we need to retry evictions? Would it ever fail due to a transient request and we wouldn't process it again?

If the eviction fails, I think the pod will continue on its backed-off crashloop anyways, so we'll get the event for the next failure.

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+                    let evict_res = pods.evict(name, &Default::default()).await;
+                    match evict_res {
+                        Ok(_) => {
+                            tracing::info!(name = format!("{namespace}/{name}"), "Evicting pod");
+                            if let Err(err) =
+                                publish_event(client.clone(), controller_pod_name.clone(), &pod)
+                                    .await
+                            {
+                                tracing::warn!(%err, name = format!("{namespace}/{name}"), "Error publishing event");
+                            }
```
mateiidavid, line 74: should we maybe process these in a separate task? if we're waiting for an answer we'll block the event loop from progressing until the future resolves.

alpeb: Great point. As talked offline, I'll try using a channel and see if I can get some backlog metrics out of it. Will also explore using a [buffered stream](https://docs.rs/futures/latest/futures/stream/struct.Buffered.html) as you suggested.

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+        let pods = kube::Api::<Pod>::namespaced(client.clone(), &namespace);
+        let evict_res = pods.evict(&name, &Default::default()).await;
+        match evict_res {
+            Ok(_) => {
+                tracing::info!(name = format!("{namespace}/{name}"), "Evicting pod");
+                metrics.evicted_pods.inc();
+                if let Err(err) =
+                    publish_k8s_event(client.clone(), controller_pod_name.clone(), &pod).await
+                {
+                    tracing::warn!(%err, name = format!("{namespace}/{name}"), "Error publishing event");
```
mateiidavid, line 124: nit: to avoid multiple string allocations, we could just declare it once as `let pod_id = format!("{namespace}/{name}");`. Or alternatively, we could do what we do in the policy controller `tracing::warn!(%err, %namespace, %name, "My msg")`.

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    let pod_evts = rt.watch_all::<Pod>(
+        watcher::Config::default()
+            .labels(DATA_PLANE_LABEL)
+            .fields(&format!("spec.nodeName={node_name}")),
+    );
+    let (tx, rx) = mpsc::channel(EVENT_CHANNEL_CAPACITY);
+    tokio::spawn(process_events(pod_evts, tx, metrics.clone()));
+
+    let client = rt.client();
+    tokio::spawn(process_pods(client, controller_pod_name, rx, metrics));
```
mateiidavid, line 42: nit: have we considered just spawning a task inside of `process_events()` for each pod we want to process? Then they can all be run in parallel by the tokio runtime. Might eliminate the need for a channel and simplify the code a little bit?

e.g.

```rust
 if terminated && !evicted {
                let namespace = pod.namespace().unwrap();
                let name = pod.name_any();
                tokio::spawn(process_pod(client.clone(), pod_name, metrics.clone());
}
```

Both the metrics and client structs should be pretty cheap to clone.

mateiidavid (commented):
This looks good to me, I still left some comments but they're nitpicky and shouldn't block. Let me know what you think :D

reinitialize-pods/src/lib.rs
alpeb: Thanks Matei, this triggered an interesting discussion (off-github). To summarize it here, spawning new tasks for each event would require us to also bound them somehow. The tokio runtime should provide us a way for doing this, but that can complicate things, so for now we're gonna keep with the channel-backed sequential approach and see how that performs.

reinitialize-pods/src/lib.rs
alpeb: Makes sense :+1:

mateiidavid (approved):
Great work @alpeb! This looks ready to ship to me.

Dockerfile-cni-plugin
```diff
@@ -16,9 +12,14 @@ ARG TARGETARCH
...
 
-##
-## Runtime
-##
+FROM --platform=$BUILDPLATFORM curlimages/curl:7.86.0 as fetch
+WORKDIR /build
+ARG TARGETARCH
+ARG LINKERD_REINITIALIZE_PODS_VERSION=v0.1.0
+# TODO: replace org with linkerd, once we make a first release
+RUN curl -LO https://github.com/alpeb/linkerd2-proxy-init/releases/download/reinitialize-pods%2F${LINKERD_REINITIALIZE_PODS_VERSION}/linkerd-reinitialize-pods-${LINKERD_REINITIALIZE_PODS_VERSION}-${TARGETARCH}.tgz
```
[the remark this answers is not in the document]
alpeb: I didn't consider a separate image given this will always be used alongside the cni plugin, so I figured I'd include it in the same image. I followed the same pattern as the network-validator, publishing a separate binary, but only when we deem necessary. But yeah, this is an additional manual step. I'll do the build here instead :+1:

reinitialize-pods/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    client: Client,
+    controller_pod_name: String,
+    mut rx: Receiver<Pod>,
+    metrics: Metrics,
+) {
+    while let Some(pod) = rx.recv().await {
+        let namespace = pod.namespace().unwrap();
+        let name = pod.name_any();
+        let pods = kube::Api::<Pod>::namespaced(client.clone(), &namespace);
+        let evict_res = pods.evict(&name, &Default::default()).await;
```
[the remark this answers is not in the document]
alpeb: Eviction is a newer API, more graceful for terminating pods, as it respects PodDisruptionBudgets and terminationGracePeriodSeconds. This doesn't strictly apply to our case where the pod is already in a bad state to begin with, but it just feels to me like a more proper way compared to a straight resource deletion.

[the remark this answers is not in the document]
olix0r: It seems likely to me that Eviction could degrade the behavior of this controller -- won't pod disruption budgets prevent these pods from being deleted as quickly as possible in some cases? Why would a user want that behavior?

This isn't strictly a blocker, but I do think it would be good to be clearer about the functional trade-offs -- it's only a proper solution if the trade-offs are desirable :)

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        watcher,
+    },
+    Client, Error, Resource, ResourceExt,
+};
+use kubert::Runtime;
+use prometheus_client::{metrics::counter::Counter, registry::Registry};
+use tokio::sync::mpsc::{self, error::TrySendError, Receiver, Sender};
+
+// ERRNO 95: Operation not supported
+pub const UNSUCCESSFUL_EXIT_CODE: i32 = 95;
```
olix0r, line 15: nit: no benefit to reuse/having this in lib.rs vs main.rs -- better to move this up to to main so it maintains locality.

olix0r, line 15: This doesn't need to be pub -- it's not used elsewhere 
```suggestion
const UNSUCCESSFUL_EXIT_CODE: i32 = 95;
```

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+                            .is_some_and(|reason| reason == CONDITION_EVICTED_REASON)
+                    })
+                })
+                .is_some();
+
+            if terminated && !evicted {
+                let namespace = pod.namespace().unwrap();
+                let name = pod.name_any();
+                // this avoids blocking the event loop
+                match tx.try_send(pod.clone()) {
```
olix0r, line 87: Note that this is going to clone all of the pod's spec. It would be substantially more efficient to only send a namespace and name over the channel.

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+                    Ok(_) => {}
+                    Err(TrySendError::Full(_)) => {
+                        tracing::warn!(%namespace, %name, "Dropped event (channel full)");
+                        metrics.queue_overflow.inc();
+                    }
+                    Err(TrySendError::Closed(_)) => {
+                        tracing::warn!(%namespace, %name, "Dropped event (channel closed or dropped)")
```
olix0r, line 94: This doesn't seem like the proper error handling semantics... If the receiver has been dropped (1) i don't think it's appropriate to log anything to the user and (2) we should stop processing events. This means nothing is interested in these updates.

cni-repair-controller/src/lib.rs
alpeb: We use the pod object down the call stack in `publish_k8s_event` to retrieve its ObjectReference. That happens to have name and namespace, so I'll pass that instead.

reinitialize-pods/src/lib.rs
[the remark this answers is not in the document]
alpeb: Ok I'll revert back to simple deletion; there isn't really a use case here for anything other than killing these pods that haven't even had a chance to start properly :+1:

alpeb:
Thanks for all the feedback @olix0r . The latest commits address all your comments, plus the tasks handling we talked about offline. And I added a unit test that also checks on the lifecycle of `process_events`.

alpeb:
Oh, also looking forward for the suggestions about the timer histogram on these tasks :wink:

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+            if terminated && !deleting {
+                let namespace = pod.namespace().unwrap();
+                let name = pod.name_any();
+                let object_ref = pod.object_ref(&());
+                // this avoids blocking the event loop
+                match tx.try_send(object_ref) {
+                    Ok(_) => {}
+                    Err(TrySendError::Full(_)) => {
+                        tracing::warn!(%namespace, %name, "Dropped event (channel full)");
```
olix0r, line 86: What's the context around these errors? Why is it safe to continue? What should future-you know about this condition? What should users know about this condition?

If understand correctly, we expect to get continuing events for all pods in a bad state? Can we document that here?

With that in mind, is this the appropriate log level and message? If this is a diagnostic for ourselves, perhaps we want to lower this to debug (we have a metric to capture during any diagnostic dump). If this is for the user: what's the proper course of action? Presumably they need to go diagnose why the other task is stuck?

Also consider that, if the queue is full, then we'll probably be in a situation where we log this *many* times (because, presumably, the rate of events is higher than we can process, so we'll remain in that state until the rate decreases). If this is for users we may want to track the states to only log when we enter this state.

With all that in mind, I think it's probably sufficient to lower the log level to debug
```suggestion
                        // If a pod is in a failed state, it should continually be
                        // reported to be in CrashLoopBackoff; so it will naturally
                        // be retried.
                        tracing::debug!(%namespace, %name, "Dropped event (channel full)");
```

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+                if let Err(err) =
+                    publish_k8s_event(client.clone(), controller_pod_name.clone(), object_ref).await
+                {
+                    tracing::warn!(%err, %namespace, %name, "Error publishing event");
+                }
+            }
+            Err(err) => {
+                tracing::warn!(%err, %namespace, %name, "Error deleting pod")
+            }
+        }
```
olix0r, line 120: Can either of these async calls block indefinitely? What bounds them?

1. We should apply a timeout to ensure this task can't get "stuck"
2. We should have metrics that count errors for each of these calls
3. We should ideally track a latency histogram for each of these calls.
4. This would be slightly cleaner with an early return :)

```suggestion
        let namespace = object_ref.namespace.clone().unwrap_or_default();
        let name = object_ref.name.clone().unwrap_or_default();
        let pods = kube::Api::<Pod>::namespaced(client.clone(), &namespace);

        let deleted = tokio::select! {
            res = pods.delete(&name, &Default::default() => res,
            _ = time::sleep(time::Duration::from_secs(1)) => {
                todo!("log/count timeout");
                continue;
            }
        };     
        if let Err(err) = deleted {
            tracing::warn!(%err, %namespace, %name, "Error deleting pod");
            todo!("count error");
            continue;
        }
        tracing::info!(%namespace, %name, "Deleted pod");
        metrics.deleted_pods.inc();
        todo!("measure latency");
       
        let event = tokio::select! {
            res = publish_k8s_event(client.clone(), controller_pod_name.clone(), object_ref => res,
            _ = time::sleep(time::Duration::from_secs(1)) => {
                todo!("log/count timeout");
                continue;
            }
        };    
        if let Err(err) = event {
            tracing::warn!(%err, %namespace, %name, "Error publishing event");            
            todo!("count error");
            continue;
        }
        todo!("measure latency");
```

cni-repair-controller/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+    let handle =
+        linkerd_cni_repair_controller::run(&mut rt, node_name, controller_pod_name, metrics);
+
+    // Block the main thread on the shutdown signal. Once it fires, wait for the background tasks to
+    // complete before exiting.
+    if rt.run().await.is_err() {
+        bail!("aborted");
+    }
+
+    handle.abort();
```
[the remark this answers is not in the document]
alpeb: Makes sense :+1:

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+    client: Client,
+    controller_pod_name: String,
+    mut rx: Receiver<ObjectReference>,
+    metrics: Metrics,
+) {
+    while let Some(object_ref) = rx.recv().await {
+        let namespace = object_ref.namespace.clone().unwrap_or_default();
+        let name = object_ref.name.clone().unwrap_or_default();
+        let pods = kube::Api::<Pod>::namespaced(client.clone(), &namespace);
+        let delete_res = pods.delete(&name, &Default::default()).await;
```
olix0r, line 106: Are the default params appropriate here?

https://docs.rs/kube/latest/kube/api/struct.DeleteParams.html

Specifically it seems like we may want to use a zero grace period.

cni-repair-controller/src/lib.rs
alpeb: The comment you added is accurate, and reflects the comment on the `EVENT_CHANNEL_CAPACITY` const. I think we'd hit this failure mostly when the rate of pod deletions is simply too high compared to this channel capacity we've set for ourselves, which I still think is a good idea. I agree in this case the user shouldn't be warned as long as the controller continues processing entries. If on the other hand the failure is caused by latency on kube-api calls in `process_pods`, the additional logs and metrics you suggest below should help diagnosing.

cni-repair-controller/src/lib.rs
alpeb: Good call :+1:

cni-repair-controller/src/lib.rs
alpeb: Agreed on all accounts. I've pushed an update addressing this and the other comments :+1:

cni-repair-controller/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+    let pod_evts = rt.watch_all::<Pod>(
+        watcher::Config::default()
+            .labels(DATA_PLANE_LABEL)
+            .fields(&format!("spec.nodeName={node_name}")),
+    );
+    let (tx, rx) = mpsc::channel(EVENT_CHANNEL_CAPACITY);
+    tokio::spawn(process_events(pod_evts, tx, metrics.clone()));
+
+    let client = rt.client();
+    tokio::spawn(process_pods(client, controller_pod_name, rx, metrics))
```
olix0r, line 58: It's probably desirable to put tracing contexts on each span for better logging/diagnostics:

```suggestion
    let (tx, rx) = mpsc::channel(EVENT_CHANNEL_CAPACITY);
    tokio::spawn(
        process_events(pod_evts, tx, metrics.clone())
            .instrument(tracing::info_span!("watch").or_current()),
    );

    let client = rt.client();
    tokio::spawn(
        process_pods(client, controller_pod_name, rx, metrics)
            .instrument(tracing::info_span!("repair").or_current()),
    )
```

olix0r (approved):

cni-repair-controller/src/lib.rs
alpeb: Sounds good :+1:

PR #316, opened 2024-01-02
Integration test for cni-repair-controller

alpeb:
(Note this will fail until linkerd/linkerd2#11699 lands)

The `integration-cni-plugin.yml` workflow (formerly known as `cni-plugin-integration.yml`) has been expanded to run the new recipe `cni-repair-controller-integration`, which performs the following steps:

- Rebuilds the `linkerd-cni-repair-controller` crate and `cni-plugin`
- Creates a new cluster at version `v1.27.6-k3s1` (version required for Calico to work)
- Triggers a new `./cni-repair-controller/integration/run.sh` script which:
  - Installs Calico
  - Installs the latest linkerd-edge CLI
  - Installs `linkerd-cni` and wait for it to become ready
  - Install the linkerd control plane in CNI mode
  - Install a `pause` DaemonSet

The `linkerd-cni` instance has been configured to include an extra initContainer that will delay its start for 15s. Since we waited for it to become ready, this doesn't affect the initial install. But then a new node is added to the cluster, and this delay allows for the new `pause` DaemonSet replica to start before the full CNI config is ready, so we can observe its failure to come up. Once the new `linkerd-cni` replica becomes ready we observe how the `pause` failed replica is replaced by a new healthy one.


cni-repair-controller/integration/linkerd-cni-config.yml
```diff
@@ -0,0 +1,7 @@
+# This config adds an extra initContainer that will make linkerd-cni to delay
+# its start for 15s, so to allow time for the pause DaemonSet to start before
+# the full CNI config is ready and enter a failure mode
+extraInitContainers:
+- name: sleep
+  image: busybox
```
mateiidavid, line 6: on the nitpicky side, the CNI plugin runs alpine, can we re-use the same image so we don't pull busybox in tests?

mateiidavid (approved):

.dockerignore
```diff
@@ -1,2 +1 @@
 rust-toolchain
-target/
```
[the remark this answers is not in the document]
alpeb: Good catch. This was a leftover from an iteration where the binary wasn't build inside the same Dockerfile.

cni-repair-controller/integration/linkerd-cni-config.yml
alpeb: Good thinking :+1:

PR #327, opened 2024-01-24
Upgrade alpine, use iptables-legacy

siggy:
PR #307 bumps alpine to 3.19.0. Unfortunately that version defaults to iptables-nft. This breaks an assumption in our Dockerfile and proxy-init that the `iptables` binary is the legacy version.

Bump alpine to 3.19.0, modify the Dockerfile to continue installing iptables-legacy, and modify the proxy-init command to default to iptables-legacy.


mateiidavid (approved):

alpeb (approved):
Wonderful! looks good to me! :+1:

PR #361, opened 2024-04-01
Expose ip tables package

zaharidichev:
In order to enable the use iptables as a library, this change moves it our of the internal package. Additionally, this PR adds functionality to remove already configured iptables.


zaharidichev:
cc @alpeb

pkg/iptables/iptables.go
```diff
@@ -99,10 +137,40 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
...
+	commands = append(commands, fc.makeDeleteChain(redirectChainName))
+
+	return commands
+}
+
 // formatComment is used to format iptables comments in such way that it is possible to identify when the rules were added.
 // This helps debug when iptables has some stale rules from previous runs, something that can happen frequently on minikube.
 func formatComment(text string) string {
-	return fmt.Sprintf("proxy-init/%s/%s", text, ExecutionTraceID)
+	return fmt.Sprintf("proxy-init/%s", text)
```
alpeb, line 173: Now that the timestamp is not included, this function comment should be amended to not make mention of "when the rules were added".

pkg/iptables/iptables.go
```diff
@@ -90,7 +91,44 @@ func ConfigureFirewall(firewallConfiguration FirewallConfiguration) error {
...
+			log.Debugf("continuing despite error: %s", err)
+		}
+	}
+
+	_, _ = executeCommand(firewallConfiguration, firewallConfiguration.makeShowAllRules())
+
+	return nil
+}
+
+// CleanupFirewallConfig removed the iptables rules that have been added as a result of
```
alpeb, line 107:
```suggestion
// CleanupFirewallConfig removes the iptables rules that have been added as a result of
```

alpeb (approved):
Looks good to me, just a couple of nits :+1:

PR #416, opened 2024-08-22
Introduce support for Projected Volumes kubernetes ServiceAccount Tokens

JJotah:
Add support for Projected Volumes kubernetes ServiceAccount Tokens. In some cloud providers we can't change the expiration token and linkerd is adding the projected volumes. So the install-cni should watch like [istio did](https://github.com/istio/istio/commit/6bbb8519b98a578cbbed657916417b7a440962db) checking the serviceaccount token

### DIFF
--- a/cni-plugin/deployment/scripts/install-cni.sh
+++ b/cni-plugin/deployment/scripts/install-cni.sh
@@ -269,18 +269,24 @@ sync() {
     # previously observed SHA (updated with each file watch) and compare it
     # against the new file's SHA. If they differ, it means something has
     # changed.
-    new_sha=$(sha256sum "${filepath}" | while read -r s _; do echo "$s"; done)
-    if [ "$new_sha" != "$prev_sha" ]; then
-      # Create but don't rm old one since we don't know if this will be configured
-      # to run as _the_ cni plugin.
-      log "New file [$filename] detected; re-installing"
-      install_cni_conf "$filepath"
+    if [[ -n "$prev_sha" ]]; then
+      new_sha=$(sha256sum "${filepath}" | while read -r s _; do echo "$s"; done)
+      if [ "$new_sha" != "$prev_sha" ]; then
+        # Create but don't rm old one since we don't know if this will be configured
+        # to run as _the_ cni plugin.
+        log "New file [$filename] detected; re-installing"
+        install_cni_conf "$filepath"
+      else
+        # If the SHA hasn't changed or we get an unrecognised event, ignore it.
+        # When the SHA is the same, we can get into infinite loops whereby a file has
+        # been created and after re-install the watch keeps triggering CREATE events
+        # that never end.
+        log "Ignoring event: $ev $filepath; no real changes detected"
+      fi
     else
-      # If the SHA hasn't changed or we get an unrecognised event, ignore it.
-      # When the SHA is the same, we can get into infinite loops whereby a file has
-      # been created and after re-install the watch keeps triggering CREATE events
-      # that never end.
-      log "Ignoring event: $ev $filepath; no real changes detected"
+      # If there is no previous SHA (as in the case of the token), reinstall cni.
+      log "Processing event for $filename with no SHA comparison needed."
+      install_cni_conf "$DEFAULT_CNI_CONF_PATH"
     fi
   fi
 }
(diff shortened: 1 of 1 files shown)


wmorgan:
Thanks @JJotah , we will take a look.

alpeb:
Fixes linkerd/linkerd2#12573

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -269,18 +269,24 @@ sync() {
...
+      fi
     else
-      # If the SHA hasn't changed or we get an unrecognised event, ignore it.
-      # When the SHA is the same, we can get into infinite loops whereby a file has
-      # been created and after re-install the watch keeps triggering CREATE events
-      # that never end.
-      log "Ignoring event: $ev $filepath; no real changes detected"
+      # If there is no previous SHA (as in the case of the token), reinstall cni.
+      log "Processing event for $filename with no SHA comparison needed."
+      install_cni_conf "$DEFAULT_CNI_CONF_PATH"
```
alpeb, line 289: We can't be sure at this point if the config is held on `$DEFAULT_CNI_CONF_PATH`, which is a `.conf` file, or a `.conflist` file.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -298,6 +304,11 @@ monitor() {
           cni_conf_sha="$(sha256sum "$directory/$filename" | while read -r s _; do echo "$s"; done)"
         fi
       fi
+    done &
+  inotifywait -m "/var/run/secrets/kubernetes.io/serviceaccount/token" -e create,delete,moved_to,modify |
```
alpeb, line 308: In `create_cni_conf()` we have `SERVICE_ACCOUNT_PATH=/var/run/secrets/kubernetes.io/serviceaccount`. Can you make that a global variable, so we can reuse it here?

alpeb (commented):
Thanks @JJotah , this makes a lot of sense!
Can I suggest introducing a variable analog to `cni_conf_sha` but tracking the sha for the token file? So we avoid introducing that new if/else in `sync()`. This would require two additional refactors though:
- Pass the entire filepath to the function instead of using `filepath="${HOST_CNI_NET}/$filename"`
- Under your new `inotifywait` call when you call `sync`, you'll have to add some logic to figure what's the actual config file name being used (as per my comment below).

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -298,6 +304,11 @@ monitor() {
           cni_conf_sha="$(sha256sum "$directory/$filename" | while read -r s _; do echo "$s"; done)"
         fi
       fi
+    done &
```
olix0r, line 307: We probably need to be careful with backgrounding these tasks -- we want to ensure that `inotifywait` failures are propagated so that the program can fail if any of these calls fail.

There are other changes we probably need to make to this script first, but we'll want to ensure that we preserve PIDs and `wait` on them to surface errors.

alpeb:
I did some more research on this, thinking it's been a while since k8s injects projected kube-api-access volumes, so why is linkerd-cni still working past the token expiration (1 hour by default)?
Inspecting the linkerd-cni pod we see the injected volume:
```yaml
  - name: kube-api-access-729gv
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
```
I could also observe how the token file got indeed recycled every hour. But actually inspecting those tokens revealed they're good for one year! So even if the token file gets renewed, linkerd-cni keeps on using the first one, which will continue to be valid.
Turns out the reason is kube-api's [flag](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#options) `--service-account-extend-token-expiration`:

> Turns on projected service account expiration extension during token generation, which helps safe transition from legacy token to bound service account token feature. If this flag is enabled, admission injected tokens would be extended up to 1 year to prevent unexpected failure during transition, ignoring value of service-account-max-token-expiration.

This flag still shows as default `true` in the docs, which explains the above. But obviously we can't rely on that, thus the importance of this PR.

alpeb:
Superseded by #440

PR #449, opened 2024-12-10
feat(linkerd-cni): add support for plain iptables commands

alpeb:
Currently the `iptables-mode` for linkerd-cni admits the values `legacy` and `nft`, which make the plugin use the `iptables-legacy[-save]` and `iptables-nft[-save]` commands respectively.

This assumes those commands are available in the node environment, given that linkerd-cni is triggered by the kubelet.

We have found that not to be the case for RHEL, where by default only `iptables[-save]` is available, which is equivalent to the `iptables-nft[-save]` command in other enviroments.

To address this case, this change adds a new possible value `iptables-mode: default` that makes the plugin use the `iptables[-save]` commands.

This has been tested successfully using RKE2 deployed in RHEL 8.10.

### DIFF
--- a/proxy-init/cmd/root.go
+++ b/proxy-init/cmd/root.go
@@ -18,15 +18,22 @@ const (
 	IPTablesModeLegacy = "legacy"
 	// IPTablesModeNFT signals the usage of the iptables-nft commands
 	IPTablesModeNFT = "nft"
+	// IPTablesModeDefault signals the usage of the iptables commands, which
+	// can be either legacy or nft
+	IPTablesModeDefault = "default"
 
-	cmdLegacy         = "iptables-legacy"
-	cmdLegacySave     = "iptables-legacy-save"
-	cmdLegacyIPv6     = "ip6tables-legacy"
-	cmdLegacyIPv6Save = "ip6tables-legacy-save"
-	cmdNFT            = "iptables-nft"
-	cmdNFTSave        = "iptables-nft-save"
-	cmdNFTIPv6        = "ip6tables-nft"
-	cmdNFTIPv6Save    = "ip6tables-nft-save"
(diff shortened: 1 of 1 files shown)


olix0r (commented):
Is the "default" mode actually the default? It looks like if the mode is unset, the default behavior is dependent on the value of FirewallBinPath, but if that's not set then we fail with an error?

My feeling is that we should not call this mode "default"--it's not the default behavior, and the default behavior can change in the future to a different mode! Is this more accurately called "infer?" Or should we explicitly call this "rhel"?

alpeb:
The `--iptables-mode` flag supersedes the `--firewall-bin-path` one. If neither is set we default to the legacy mode (`iptables-legacy[-save]`). 
You're right, I was hesitant about using "default" for the new value. I wouldn't use "rhel" as this behavior might be found elsewhere. "infer" doesn't ring a bell either because we're not deducing about one command or another... Other options that come to mind: bare, flat, plain.
Let me know what you think.

olix0r:
"plain" sounds okay to me ¯\_(ツ)_/¯

olix0r (approved):

alpeb:
Currently the `iptables-mode` for linkerd-cni admits the values `legacy` and `nft`, which make the plugin use the `iptables-legacy[-save]` and `iptables-nft[-save]` commands respectively.

This assumes those commands are available in the node environment, given that linkerd-cni is triggered by the kubelet.

We have found that not to be the case for RHEL, where by default only `iptables[-save]` is available, which is equivalent to the `iptables-nft[-save]` command in other enviroments.

To address this case, this change adds a new possible value `iptables-mode: plain` that makes the plugin use the `iptables[-save]` commands.

This has been tested successfully using RKE2 deployed in RHEL 8.10.

PR #477, opened 2025-02-15
BUG: fix race condition

sdickhoven:


cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -239,12 +239,12 @@ sync() {
 
   local config_file_count
   local new_sha
-  if [ "$ev" = 'CREATE' ] || [ "$ev" = 'MOVED_TO' ] || [ "$ev" = 'MODIFY' ]; then
+  if [ "$ev" = 'CREATE' -o "$ev" = 'MOVED_TO' -o "$ev" = 'MODIFY' ]; then
```
sdickhoven, line 242: unrelated. just more efficient.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -239,12 +239,12 @@ sync() {
...
     # When the event type is 'CREATE', 'MOVED_TO' or 'MODIFY', we check the
     # previously observed SHA (updated with each file watch) and compare it
     # against the new file's SHA. If they differ, it means something has
     # changed.
-    new_sha=$(sha256sum "${filepath}" | while read -r s _; do echo "$s"; done)
+    new_sha=$(sha256sum "${filepath}" | awk '{print $1}')
```
sdickhoven, line 247: unrelated. just more efficient/concise.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -271,7 +271,7 @@ monitor_cni_config() {
         sync "$filename" "$action" "$cni_conf_sha"
         # calculate file SHA to use in the next iteration
         if [[ -e "$directory/$filename" ]]; then
-          cni_conf_sha="$(sha256sum "$directory/$filename" | while read -r s _; do echo "$s"; done)"
+          cni_conf_sha=$(sha256sum "$directory/$filename" | awk '{print $1}')
```
sdickhoven, line 274: unrelated. just more efficient/concise.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -310,32 +310,29 @@ rm -f "${DEFAULT_CNI_CONF_PATH}"
...
+# Otherwise, new CNI config files can be created just _after_ the initial round
+# of patching and just _before_ we set up the `inotifywait` loop to detect new
+# CNI config files.
+cni_conf_sha="__init__"
+monitor_cni_config &
+
 # Append our config to any existing config file (*.conflist or *.conf)
 config_files=$(find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \))
-if [ -z "$config_files" ]; then
-    log "No active CNI configuration files found"
```
sdickhoven, line 316: this `if` condition is already covered by the `if` condition below.

i.e. if the variable `config_files` is an empty string (`-z`) then

```bash
echo "$config_files" | grep -v linkerd | sort | wc -l
```

will be `0`... and the next `if` condition already checks for that.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -310,32 +310,29 @@ rm -f "${DEFAULT_CNI_CONF_PATH}"
 
 install_cni_bin
 
+# The CNI config monitor must be set up _before_ we start patching CNI config
+# files!
+# Otherwise, new CNI config files can be created just _after_ the initial round
+# of patching and just _before_ we set up the `inotifywait` loop to detect new
+# CNI config files.
+cni_conf_sha="__init__"
+monitor_cni_config &
```
sdickhoven, line 319: this is the crux of my pr.

if `monitor_cni_config()` is not started _before_ we go looking for cni config files to patch, then it is possible for cni config files to be created _after_ the `find` command below is done generating the list of cni config files that have to be patched, but _before_ the `inotifywait` in `monitor_cni_config()` is up and watching for new files.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -310,32 +310,29 @@ rm -f "${DEFAULT_CNI_CONF_PATH}"
...
-if [[ -n "$conf" ]]; then
-  cni_conf_sha="$(sha256sum "$conf" | while read -r s _; do echo "$s"; done)"
+  find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
+    while read -r -d $'\0' file; do
+      log "Trigger CNI config detection for $file"
+      tmp_file="$(mktemp -u /tmp/linkerd-cni.patch-candidate.XXXXXX)"
+      cp -fp "$file" "$tmp_file"
+      # The following will trigger the `sync()` function via `inotifywait` in
+      # `monitor_cni_config()`.
+      mv -f "$tmp_file" "$file"
```
sdickhoven, line 334: rather than patching cni config files here, we are going to trigger the `inotifywait` in the (already up and running) `monitor_cni_config()` which will call `sync()` to do the actual patching.

cni-plugin/deployment/scripts/install-cni.sh
sdickhoven: there is no reason for `cni_conf_sha` to have an actual value at this point. we could also do

```bash
cni_conf_sha=
```

here. the only important thing is that whatever we set the initial value to can't possibly be a valid `sha256sum`.

cni-plugin/deployment/scripts/install-cni.sh
sdickhoven: this has the pleasant side effect of consolidating the patching logic to _one_ place inside this script.

cni-plugin/deployment/scripts/install-cni.sh
sdickhoven: also removing the `"` around `$()` since they are not needed and because that's consistent with how it's done above.

cni-plugin/deployment/scripts/install-cni.sh
sdickhoven: by the way, i'm not quite sure if `sync()` is idempotent. i.e. if it's smart enough to **not** patch a cni config file that has _already_ been patched.

if not, then we should insert something like

```bash
grep -qw linkerd "$file" && continue || echo "Found unpatched CNI config file: $file"
```

at the top of the while loop.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -310,34 +310,35 @@ rm -f "${DEFAULT_CNI_CONF_PATH}"
...
-        log "Installing CNI configuration for $file"
-        create_kubeconfig
-        create_cni_conf
-        install_cni_conf "$file"
+        log "Trigger CNI config detection for $file"
+        tmp_file="$(mktemp -u /tmp/linkerd-cni.patch-candidate.XXXXXX)"
+        cp -fp "$file" "$tmp_file"
+        # The following will trigger the `sync()` function via `inotifywait` in
+        # `monitor_cni_config()`.
+        mv -f "$tmp_file" "$file"
```
sdickhoven, line 337: instead of running

```bash
create_kubeconfig
create_cni_conf
install_cni_conf "$file"
```

we're simply going to copy the cni config file to a temporary location and then move it back to trigger the `inotifywait` in `monitor_cni_config()`.

this will then call `sync()` which takes care of the patching.

sdickhoven: by the way, it is important for `sync()` to be idempotent. i.e. it must be smart enough to **not** add a second linkerd block to a cni config file that already has one.

otherwise, there would _still_ be a race condition here since the `sha256sum` logic cannot be relied upon when dealing with more than _one_ cni config file.

based on my testing, idempotence is achieved by the logic in [filter.jq](https://github.com/linkerd/linkerd2-proxy-init/blob/cni-plugin/v1.6.0/cni-plugin/deployment/scripts/filter.jq). ✅

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -310,34 +310,35 @@ rm -f "${DEFAULT_CNI_CONF_PATH}"
...
 fi
 
-# Compute SHA for first config file found; this will be updated after every iteration.
-# First config file is likely to be chosen as the de facto CNI config by the
-# host.
-conf="$(find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) | sort | head -n 1)"
-cni_conf_sha=""
-if [[ -n "$conf" ]]; then
-  cni_conf_sha="$(sha256sum "$conf" | while read -r s _; do echo "$s"; done)"
-fi
```
sdickhoven, line 339: superfluous.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -310,34 +310,35 @@ rm -f "${DEFAULT_CNI_CONF_PATH}"
...
     log "No active CNI configuration files found"
   else
     find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
       while read -r -d $'\0' file; do
-        log "Installing CNI configuration for $file"
-        create_kubeconfig
-        create_cni_conf
-        install_cni_conf "$file"
+        log "Trigger CNI config detection for $file"
+        tmp_file="$(mktemp -u /tmp/linkerd-cni.patch-candidate.XXXXXX)"
```
sdickhoven, line 333: by the way

```bash
mktemp -u /tmp/linkerd-cni.patch-candidate.XXXXXX
```

gives the same result as

```bash
mktemp -ut linkerd-cni.patch-candidate.XXXXXX
```

sdickhoven:
hi @alpeb 👋 

i have taken out the unrelated changes. ✅ 

happy to submit another pr for those unrelated changes but they are 100% cosmetic and 0% functional. so there's really no need.

sdickhoven:
by the way, i just thought of a way in which `install-cni.sh` can go into an infinite patching loop.

specifically, the current `sha256sum` logic assumes that there is only ever going to be _one_ cni config file.

picture this scenario: two new cni config files are created at very nearly the same time.

let's call them

* `cni1.conflist`
* `cni2.conflist`

`sync()` updates `cni1.conflist` and sets `cni_conf_sha` to the sha sum for _that_ file.

moving the updated `cni1.conflist` file into place triggers another inotifywait event.

but before we can deal with _that_ inotifywait event, we first have to deal with the one that is already waiting in the queue for `cni2.conflist`...

`sync()` updates `cni2.conflist` and sets `cni_conf_sha` to the sha sum for _that_ file.

moving the updated `cni2.conflist` file into place triggers another inotifywait event.

but before we can deal with _that_ inotifywait event, we first have to deal with the one that is already waiting in the queue for `cni1.conflist`...

...and since `cni_conf_sha` now contains the sha sum of `cni2.conflist`, `sync()` will not detect that it has already updated `cni1.conflist`...

so the cycle repeats... forever.

should i fix this behavior in this pr or create a new pr?

it is technically a _different_ error case (not a race condition but an infinite loop).

i would address this by updating the sha sum logic to store _all_ sha sums in a variable so that the sha sum for the correct file can be looked up.

sdickhoven:
this is what my fix for the infinite patching loop would look like (using this pr as base):

```diff
diff --git a/cni-plugin/deployment/scripts/install-cni.sh b/cni-plugin/deployment/scripts/install-cni.sh
index cdf166c..b829314 100755
--- a/cni-plugin/deployment/scripts/install-cni.sh
+++ b/cni-plugin/deployment/scripts/install-cni.sh
@@ -268,10 +268,11 @@ monitor_cni_config() {
     while read -r directory action filename; do
       if [[ "$filename" =~ .*.(conflist|conf)$ ]]; then 
         log "Detected change in $directory: $action $filename"
-        sync "$filename" "$action" "$cni_conf_sha"
+        sync "$filename" "$action" "$(jq -r --arg file "$filename" '.[$file]' <<< "$cni_conf_sha")"
         # calculate file SHA to use in the next iteration
         if [[ -e "$directory/$filename" ]]; then
-          cni_conf_sha="$(sha256sum "$directory/$filename" | while read -r s _; do echo "$s"; done)"
+          new_sha="$(sha256sum "$directory/$filename" | while read -r s _; do echo "$s"; done)"
+          cni_conf_sha="$(jq -c --arg file "$filename" --arg sha "$new_sha" '. * {$file: $sha}' <<< "$cni_conf_sha")"
         fi
       fi
     done
@@ -315,7 +316,7 @@ install_cni_bin
 # Otherwise, new CNI config files can be created just _after_ the initial round
 # of patching and just _before_ we set up the `inotifywait` loop to detect new
 # CNI config files.
-cni_conf_sha="__init__"
+cni_conf_sha='{}'
 monitor_cni_config &
 
 # Append our config to any existing config file (*.conflist or *.conf)
```

alpeb:
Sorry for the late reply... As you correctly realized, this script was designed supposing the existence of a single cni config file. I'd like to spend more time investigating how linkerd-cni should behave in the presence of multiple such files before evaluating your solution. So let's for now leave your last diff posted as a comment out of the current PR. I'll give another look and round of tests to the current commits in the following days, so we can move forward.

sdickhoven:
>Given that we're now calling monitor_cni_config & earlier, the entire block that follows it should no longer be necessary, right?

which block are you referring to?

this?:

```bash
# Append our config to any existing config file (*.conflist or *.conf)
config_files=$(find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \))
if [ -z "$config_files" ]; then
    log "No active CNI configuration files found"
else
  config_file_count=$(echo "$config_files" | grep -v linkerd | sort | wc -l)
  if [ "$config_file_count" -eq 0 ]; then
    log "No active CNI configuration files found"
  else
    find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
      while read -r -d $'\0' file; do
        log "Installing CNI configuration for $file"
...
```

this block is _definitely_ needed!

because cni config files can already exist _before_ linkerd-cni even starts up. and you would never get an event from `inotifywait` for those files that already exist in `/etc/cni/net.d`.

the logic is:

1. set up `inotifywait`
2. _force_ any already existing files to trigger `inotifywait`

i guess that the _change_ in the block that follows `monitor_cni_config &` is not strictly necessary (because all that really matters is that the script dosn't _miss_ new cni config files).

but i figured that consolidating the patching logic made sense.

does that answer your question? apologies if i did not understand your question correctly.

the log output looks exactly like what i would expect from my changes.

sdickhoven:
i hope this answers your question. please let me know if i'm not making sense.

sdickhoven:
any progress on this?

cni chaining is kind of a thing and it would be good for linkerd to work correctly in environments where cni chaining is already in place... i.e. where linkerd isn't the _only_ chained cni plugin.

chaining cilium to aws vpc cni is definitely a pretty common setup based on what i see in the cilium slack workspace.

let me know if there's anything else i can do to help get this very real race condition fixed.

alpeb:
Hi, thanks for pinging back; looking again into this...
You're right about requiring the logic to be triggered for the initial file. Quick question: is there a more atomic way to trigger that than having the copy followed by the move? Would just calling `touch` on the file trigger the sync? (that'll likely require adding another event to the list of inotifywait events).

As a side note, I've just realized libcni [recently released](https://github.com/containernetworking/cni/releases/tag/v1.3.0) a way of doing safe subdirectory-based plugin config loading, which would save us from all this config patching nonsense... but that approach will have to wait till it gets picked up by the major cloud providers. Perhaps we could have users opt into that behavior. Not asking you to implement any of that, just wanted to share what I found out :slightly_smiling_face:

sdickhoven:
>is there a more atomic way to trigger that than having the copy followed by the move?

yes. good point. `touch`ing a file would be cleaner indeed. something like this should work:

```
root@5a06a8dd2c88:~# inotifywait -m . -e moved_to,close_write &
[1] 211
Setting up watches.
Watches established.

root@5a06a8dd2c88:~# touch ttt
./ CLOSE_WRITE,CLOSE ttt
root@5a06a8dd2c88:~# echo > abc
./ CLOSE_WRITE,CLOSE abc
root@5a06a8dd2c88:~# cat ttt
root@5a06a8dd2c88:~# echo > def
./ CLOSE_WRITE,CLOSE def
root@5a06a8dd2c88:~# touch def
./ CLOSE_WRITE,CLOSE def
root@5a06a8dd2c88:~# touch /tmp/ttt
root@5a06a8dd2c88:~# mv /tmp/ttt .
./ MOVED_TO ttt
root@5a06a8dd2c88:~# rm ttt
root@5a06a8dd2c88:~# rm abc
root@5a06a8dd2c88:~#
```

>I've just realized libcni [recently released](https://github.com/containernetworking/cni/releases/tag/v1.3.0) a way of doing safe subdirectory-based plugin config loading, which would save us from all this config patching nonsense

😍 that's great! i hope it becomes available soon. that would certainly make things easier.

do you want me to update my pr with the `touch` logic (which requires an update the `inotifywait` events we want to watch for)?

also, do you want me to open another pr for the other race condition i pointed out about the infinite patching loop due to the flawed shasum logic?

if so, then i'd want to wait until this pr is merged before i submit the second pr.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -321,23 +329,14 @@ else
...
     find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
       while read -r -d $'\0' file; do
-        log "Installing CNI configuration for $file"
-        create_kubeconfig
-        create_cni_conf
-        install_cni_conf "$file"
+        log "Trigger CNI config detection for $file"
+        # The following will trigger the `sync()` function via `inotifywait` in
+        # `monitor_cni_config()`.
+        touch "$file"
```
sdickhoven, line 335: fyi: there may be filesystems out there that round the `mtime` attribute to seconds.

see https://github.com/kubernetes/minikube/pull/1594

this may render a `touch` ineffective (because the old and new `mtime` could end up being the _same_ which may then **not** trigger an `ATTRIB` event) but i have no way of knowing/testing that.

the _safest_ / most universally compatible way to do this is probably what i had before. but this is definitely "cleaner".

up to you if you want to keep the `touch` logic or revert back to the `mv` logic. happy to revert if you tell me to.

sdickhoven:
>but... i don't want to talk you into doing something that you're not comfortable with. so... your call.

also... no rush. ☺️ 

we currently have the workaround outlined in the pr description deployed in all of our clusters. so _we_ are fine. ✅ 

if you'd rather rewrite the cni patch logic in go or wait for the new subdirectory-based plugin config, that's fine by us.

but other linkerd-cni users may run into this race condition and won't be able to effectively troubleshoot / mitigate it. 🤷

sdickhoven:
so... i found a number of additional race conditions in this script.

none of them have to do with bash but with the fact that you are dealing with a filesystem. so you have to really think about [atomicity](https://en.wikipedia.org/wiki/Linearizability).

anyway... this is a diff based on the _current_ pr of all the changes that i would _like_ to submit in order to remove _all_ race conditions (that i have discovered while refactoring):

```diff
diff --git a/cni-plugin/deployment/scripts/install-cni.sh b/cni-plugin/deployment/scripts/install-cni.sh
index fe2ffa0..ce4a4c4 100755
--- a/cni-plugin/deployment/scripts/install-cni.sh
+++ b/cni-plugin/deployment/scripts/install-cni.sh
@@ -77,7 +77,7 @@ cleanup() {
   # Find all conflist files and print them out using a NULL separator instead of
   # writing each file in a new line. We will subsequently read each string and
   # attempt to rm linkerd config from it using jq helper.
-  local cni_data=''
+  local cni_data
   find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' \) -print0 |
     while read -r -d $'\0' file; do
       log "Removing linkerd-cni config from $file"
@@ -189,91 +189,100 @@ EOF
   # Use alternative command character "~", since these include a "/".
   sed -i s~__KUBECONFIG_FILEPATH__~"${DEST_CNI_NET_DIR}/${KUBECONFIG_FILE_NAME}"~g ${TMP_CONF}
 
-  log "CNI config: $(cat ${TMP_CONF})"
+  log "CNI config: $(cat $TMP_CONF)"
 }
 
 install_cni_conf() {
   local cni_conf_path=$1
 
-  local tmp_data=''
-  local conf_data=''
-  if [ -e "${cni_conf_path}" ]; then
-   # Add the linkerd-cni plugin to the existing list
-   tmp_data=$(cat "${TMP_CONF}")
-   conf_data=$(jq --argjson CNI_TMP_CONF_DATA "${tmp_data}" -f /linkerd/filter.jq "${cni_conf_path}")
-   echo "${conf_data}" > ${TMP_CONF}
-  fi
+  local tmp_data=$(cat "$TMP_CONF")
+  local conf_data=$(jq --argjson CNI_TMP_CONF_DATA "$tmp_data" -f /linkerd/filter.jq "$cni_conf_path" || true)
+
+  # Ensure that CNI config file did not disappear during processing.
+  [ -n "$conf_data" ] || return
+
+  # Add the linkerd-cni plugin to the existing list.
+  echo "$conf_data" > "$TMP_CONF"
 
-  # If the old config filename ends with .conf, rename it to .conflist, because it has changed to be a list
-  filename=${cni_conf_path##*/}
-  extension=${filename##*.}
+  # If the old config filename ends with .conf, rename it to .conflist because
+  # it has changed to be a list.
+  local filename=${cni_conf_path##*/}
+  local extension=${filename##*.}
   # When this variable has a file, we must delete it later.
   old_file_path=
-  if [ "${filename}" != '01-linkerd-cni.conf' ] && [ "${extension}" = 'conf' ]; then
-   old_file_path=${cni_conf_path}
-   log "Renaming ${cni_conf_path} extension to .conflist"
-   cni_conf_path="${cni_conf_path}list"
+  if [ "$filename" != '01-linkerd-cni.conf' ] && [ "$extension" = 'conf' ]; then
+    old_file_path=$cni_conf_path
+    log "Renaming $cni_conf_path extension to .conflist"
+    cni_conf_path="${cni_conf_path}list"
   fi
 
   # Move the temporary CNI config into place.
-  mv "${TMP_CONF}" "${cni_conf_path}" || exit_with_error 'Failed to mv files.'
-  [ -n "$old_file_path" ] && rm -f "${old_file_path}" && log "Removing unwanted .conf file"
+  mv "$TMP_CONF" "$cni_conf_path" || exit_with_error 'Failed to mv files.'
+  [ -n "$old_file_path" ] && rm -f "$old_file_path" && log "Removing unwanted .conf file"
 
-  log "Created CNI config ${cni_conf_path}"
+  log "Created CNI config $cni_conf_path"
 }
 
-# Sync() is responsible for reacting to file system changes. It is used in
-# conjunction with inotify events; sync() is called with the name of the file
-# that has changed, the event type (which can be either 'CREATE', 'DELETE',
-# 'MOVED_TO' or 'MODIFY', and the previously observed SHA of the configuration
-# file.
+# `sync()` is responsible for reacting to file system changes. It is used in
+# conjunction with inotify events; `sync()` is called with the event type (which
+# can be either 'CREATE', 'DELETE', 'MOVED_TO', 'MODIFY' or 'DELETE') and the
+# name of the file that has changed 
 #
-# Based on the changed file and event type, sync() might re-install the CNI
+# Based on the changed file and event type, `sync()` might re-install the CNI
 # plugin's configuration file.
 sync() {
-  local filename=$1
-  local ev=$2
-  local filepath="${HOST_CNI_NET}/$filename"
-
-  local prev_sha=$3
-
-  local config_file_count
-  local new_sha
-  if [ "$ev" = 'CREATE' ] || [ "$ev" = 'MOVED_TO' ] || [ "$ev" = 'MODIFY' ] || [ "$ev" = 'ATTRIB' ]; then
-    # When the event type is 'CREATE', 'MOVED_TO' or 'MODIFY', we check the
-    # previously observed SHA (updated with each file watch) and compare it
-    # against the new file's SHA. If they differ, it means something has
-    # changed.
-    new_sha=$(sha256sum "${filepath}" | while read -r s _; do echo "$s"; done)
-    if [ "$new_sha" != "$prev_sha" ]; then
-      # Create but don't rm old one since we don't know if this will be configured
-      # to run as _the_ cni plugin.
-      log "New/changed file [$filename] detected; re-installing"
-      create_kubeconfig
-      create_cni_conf
-      install_cni_conf "$filepath"
-    else
-      # If the SHA hasn't changed or we get an unrecognised event, ignore it.
-      # When the SHA is the same, we can get into infinite loops whereby a file has
-      # been created and after re-install the watch keeps triggering CREATE events
-      # that never end.
-      log "Ignoring event: $ev $filepath; no real changes detected"
-    fi
+  local ev=$1
+  local file=$2
+
+  [[ "$file" =~ .*.(conflist|conf)$ ]] || return
+
+  log "Detected event: $ev $file"
+
+  # The following covers the case where a CNI config file is created and then
+  # immediately removed while `install_cni_conf()` is patching it.
+  # Without this, `install_cni_conf()` could create a patched CNI config file
+  # that should not exist.
+  if [ "$ev" = "DELETE" ]; then
+    rm -f "$file"
+    return
+  fi
+
+  # Retrieve previous SHA of detected file (if any) and compute current SHA.
+  local previous_sha=$(jq -r --arg f "$file" '.[$f] | select(.)' <<< "$cni_conf_sha")
+  local current_sha=$( (sha256sum "$file" || true) | awk '{print $1}' )
+
+  if [ -n "$current_sha" ] && [ "$current_sha" != "$previous_sha" ]; then
+    log "New/changed file [$file] detected; re-installing"
+    create_kubeconfig
+    create_cni_conf
+    install_cni_conf "$file"
+
+    # Store SHA of each patched file in global `cni_conf_sha` variable.
+    #
+    # The below logic assumes that the `cni_conf_sha` variable is already a
+    # valid JSON object. So this variable must be initialized with '{}'!
+    #
+    # E.g. (pretty-printed; actual variable stores compact JSON object)
+    # {
+    #   "/etc/cni/net.d/05-foo.conflist": "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c",
+    #   "/etc/cni/net.d/10-bar.conflist": "7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730"
+    # }
+    local new_sha=$( (sha256sum "$file" || true) | awk '{print $1}' )
+    cni_conf_sha=$(jq -c --arg f "$file" --arg sha "$new_sha" '. * {$f: $sha}' <<< "$cni_conf_sha")
+  else
+    # If the SHA hasn't changed or the detected file has disappeared, ignore it.
+    # When the SHA is the same, we can get into infinite loops whereby a file
+    # has been created and after re-install the watch keeps triggering MOVED_TO
+    # events that never end.
+    log "Ignoring event: $ev $file; no real changes detected or file disappeared"
   fi
 }
 
 # monitor_cni_config starts a watch on the host's CNI config directory
 monitor_cni_config() {
-  inotifywait -m "${HOST_CNI_NET}" -e create,moved_to,modify,attrib |
+  inotifywait -m "${HOST_CNI_NET}" -e create,moved_to,modify,delete |
     while read -r directory action filename; do
-      if [[ "$filename" =~ .*.(conflist|conf)$ ]]; then 
-        log "Detected change in $directory: $action $filename"
-        sync "$filename" "$action" "$cni_conf_sha"
-        # calculate file SHA to use in the next iteration
-        if [[ -e "$directory/$filename" ]]; then
-          cni_conf_sha="$(sha256sum "$directory/$filename" | while read -r s _; do echo "$s"; done)"
-        fi
-      fi
+      sync "$action" "$directory/$filename"
     done
 }
 
@@ -284,16 +293,16 @@ monitor_cni_config() {
 # only reacting to direct creation of a "token" file, or creation of
 # directories containing a "token" file.
 monitor_service_account_token() {
-    inotifywait -m "${SERVICEACCOUNT_PATH}" -e create |
-      while read -r directory _ filename; do
-        target=$(realpath "$directory/$filename")
-        if [[ (-f "$target" && "${target##*/}" == "token") || (-d "$target" && -e "$target/token") ]]; then
-          log "Detected creation of file in $directory: $filename; recreating kubeconfig file"
-          create_kubeconfig
-        else
-          log "Detected creation of file in $directory: $filename; ignoring"
-        fi
-      done
+  inotifywait -m "${SERVICEACCOUNT_PATH}" -e create |
+    while read -r directory _ filename; do
+      target=$(realpath "$directory/$filename")
+      if [[ (-f "$target" && "${target##*/}" == "token") || (-d "$target" && -e "$target/token") ]]; then
+        log "Detected creation of file in $directory: $filename; recreating kubeconfig file"
+        create_kubeconfig
+      else
+        log "Detected creation of file in $directory: $filename; ignoring"
+      fi
+    done
 }
 
 log() {
@@ -315,26 +324,23 @@ install_cni_bin
 # Otherwise, new CNI config files can be created just _after_ the initial round
 # of patching and just _before_ we set up the `inotifywait` loop to detect new
 # CNI config files.
-cni_conf_sha="__init__"
+cni_conf_sha='{}'
 monitor_cni_config &
 
 # Append our config to any existing config file (*.conflist or *.conf)
-config_files=$(find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \))
+config_files=$(find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) | grep -v linkerd)
 if [ -z "$config_files" ]; then
-    log "No active CNI configuration files found"
+  log "No active CNI configuration files found"
 else
-  config_file_count=$(echo "$config_files" | grep -v linkerd | sort | wc -l)
-  if [ "$config_file_count" -eq 0 ]; then
-    log "No active CNI configuration files found"
-  else
-    find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
-      while read -r -d $'\0' file; do
-        log "Trigger CNI config detection for $file"
-        # The following will trigger the `sync()` function via `inotifywait` in
-        # `monitor_cni_config()`.
-        touch "$file"
-      done
-  fi
+  find "${HOST_CNI_NET}" -maxdepth 1 -type f \( -iname '*conflist' -o -iname '*conf' \) -print0 |
+    while read -r -d $'\0' file; do
+      log "Trigger CNI config detection for $file"
+      tmp_file="$(mktemp -u /tmp/linkerd-cni.patch-candidate.XXXXXX)"
+      cp -fp "$file" "$tmp_file"
+      # The following will trigger the `sync()` function via `inotifywait` in
+      # `monitor_cni_config()`.
+      mv -f "$tmp_file" "$file"
+    done
 fi
 
 # Watch in bg so we can receive interrupt signals through 'trap'. From 'man
```

anyway, as you can see, it's quite a comprehensive change.

ps: the above has not been tested yet. i wrote it "blind". if you'd be ok with such a comprehensive change i would be happy to test this exhaustively (which will probably take about half a day).

and, if you want, i could also write unit tests to really exercise all the different race condition edge cases.

but that'll be a bit of a project and i want to make sure that you'd be open to making such a comprehensive change (assuming that i can prove to you that it is functionally equivalent to what you have now but the race conditions removed (and more consistent formatting).

sdickhoven:
>

e.g. [this](https://github.com/sdickhoven/linkerd2-proxy-init/blob/812d8b5/cni-plugin/deployment/scripts/install-cni.sh#L200-L205) is a race condition.

if the cni config file is removed between

```bash
if [ -e "${cni_conf_path}" ]; then
```

and

```bash
conf_data=$(jq --argjson CNI_TMP_CONF_DATA "${tmp_data}" -f /linkerd/filter.jq "${cni_conf_path}")
```

the script will exit and linkerd-cni will be permanently "disabled".

as you can see, i am doing the following instead which is robust against this edge case:

```bash
conf_data=$(jq --argjson CNI_TMP_CONF_DATA "$tmp_data" -f /linkerd/filter.jq "$cni_conf_path" || true)
[ -n "$conf_data" ] || return
```

sdickhoven:
anyway, most people are familiar with using locks/mutexes when dealing with concurrent access to mutable data... but if you _cannot_ do that (as is the case here), most engineers will probably struggle to write "correct" code.

so libcni's [recently released](https://github.com/containernetworking/cni/releases/tag/v1.3.0) mechanism for subdirectory-based plugin config loading definitely addresses a _real_ challenge.

sdickhoven:
i just did a bit of testing and i found even more race conditions. i'll update the above diff once i'm happy with what i have.

sdickhoven:
ok. i have pushed all of the changes that i strongly believe _should_ be made to eliminate all of the race conditions that i have discovered... except for one highly improbably one (that i have documented) that i was unable to address.

i *CANNOT* make this break even when i hammer it incessantly with:

```
while :
do
  echo '{"cniVersion": "0.4.0", "name": "aws-cni", "disableCheck": true, "plugins": [{"name": "foo"}]}' > tmp/01-foo.conflist
  echo '{"cniVersion": "0.4.0", "name": "aws-cni", "disableCheck": true, "plugins": [{"name": "bar"}]}' > tmp/02-bar.conflist
  mv tmp/01-foo.conflist etc/cni/net.d/01-foo.conflist
  mv tmp/02-bar.conflist etc/cni/net.d/02-bar.conflist
  rm tmp/02-bar.conflist etc/cni/net.d/02-bar.conflist
done
```

my test environment for this was:

```
$ docker run -ti --rm \
    -v $(pwd)/serviceaccount:/var/run/secrets/kubernetes.io/serviceaccount \
    -v $(pwd)/etc/cni/net.d:/host/etc/cni/net.d \
    -v $(pwd)/opt/cni/bin:/host/opt/cni/bin \
    -v $(pwd)/tmp:/tmp \
    --mount type=bind,src=$HOME/git/contrib/linkerd2-proxy-init/cni-plugin/deployment/scripts/install-cni.sh,dst=/linkerd/install-cni.sh,ro \
    -e KUBERNETES_SERVICE_HOST=::1 \
    -e KUBERNETES_SERVICE_PORT=443 \
    -e SKIP_TLS_VERIFY=true \
    -e CNI_NETWORK_CONFIG='{"name":"linkerd-cni","type":"linkerd-cni","log_level":"info","kubernetes":{"kubeconfig":"__KUBECONFIG_FILEPATH__"}}' \
    -e SLEEP=true \
    cr.l5d.io/linkerd/cni-plugin:v1.6.2
```

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -176,104 +176,113 @@ create_cni_conf() {
...
+  # If the SHA hasn't changed or the detected file has disappeared, ignore it.
+  # When the SHA is the same, we can get into infinite loops whereby a file
+  # has been created and after re-install the watch keeps triggering MOVED_TO
+  # events that never end.
+  # If the `current_sha` variable is blank then the detected CNI config file has
+  # disappeared and no further action is required.
+  # There exists an unhandled (highly improbable) edge case where a CNI plugin
+  # creates a config file and then _immediately_ removes it again _while_ we are
+  # in the process of patching it. If this happens, we may create a patched CNI
+  # config file that should *not* exist.
```
sdickhoven, line 270: addressing this race condition is difficult because it would require us to monitor and respond to `DELETE` events.

however, i have seen those events get emitted when this script does a `mv` to overwrite the original cni config file... and i can't think of a way to distinguish between _genuine_ `DELETE` events and spurious ones.

this may just be an artifact of my container runtime (docker in this case) but we obviously want to make sure that our logic works for _all_ container runtimes, n'est-ce pas?

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -306,36 +315,32 @@ log() {
...
+rm -f "$DEFAULT_CNI_CONF_PATH"
 
 install_cni_bin
 
+# The CNI config monitor must be set up _before_ we start patching existing CNI
+# config files!
+# Otherwise, new CNI config files can be created just _after_ the initial round
+# of patching and just _before_ we set up the `inotifywait` loop to detect new
+# CNI config files.
+CNI_CONF_SHA='{}'
```
sdickhoven, line 327: it's customary to use all caps for (explicitly) global variables.

cni-plugin/deployment/scripts/install-cni.sh
```diff
@@ -176,104 +176,113 @@ create_cni_conf() {
...
+  # Add the linkerd-cni plugin to the existing list.
+  echo "$conf_data" > "$TMP_CONF"
 
-  # If the old config filename ends with .conf, rename it to .conflist, because it has changed to be a list
-  filename=${cni_conf_path##*/}
-  extension=${filename##*.}
+  # If the old config filename ends with .conf, rename it to .conflist because
+  # it has changed to be a list.
+  local filename=${cni_conf_path##*/}
+  local extension=${filename##*.}
```
sdickhoven, line 210: these are rather generic variable names, so it's safest to make them `local` in scope.

alpeb:
Hey @sdickhoven , I'm back at looking into this. Do you mind moving all the changes not strictly related to the matter at hand (race conditions) into a separate PR?

sdickhoven:
> Hey @sdickhoven , I'm back at looking into this. Do you mind moving all the changes not strictly related to the matter at hand (race conditions) into a separate PR?

done ✅

alpeb:
Actually you might be forced to force-push to fix the DCO :wink:

sdickhoven:
sorry i just saw

>Commit sha: [812d8b5](https://github.com/linkerd/linkerd2-proxy-init/pull/477/commits/812d8b50e1357edce066b8889f9eddf4dd1535a9), Author: Simon Dickhoven, Committer: Simon Dickhoven; The sign-off is missing.
>Commit sha: [defda27](https://github.com/linkerd/linkerd2-proxy-init/pull/477/commits/defda271cfcec570e351c157bada73da5112a693), Author: Simon Dickhoven, Committer: Simon Dickhoven; The sign-off is missing.
>Commit sha: [f6d3844](https://github.com/linkerd/linkerd2-proxy-init/pull/477/commits/f6d3844cc77265d267b4e35a0b8a453f1dbce51a), Author: Simon Dickhoven, Committer: Simon Dickhoven; The sign-off is missing.
>Commit sha: [4112de0](https://github.com/linkerd/linkerd2-proxy-init/pull/477/commits/4112de07d4c739d4ea9fb2ed6f8dbb0bb642fffa), Author: Simon Dickhoven, Committer: Simon Dickhoven; The sign-off is missing.
>Commit sha: [335bb3b](https://github.com/linkerd/linkerd2-proxy-init/pull/477/commits/335bb3b41116b49118be68d80b27e73fbd897cd2), Author: Simon Dickhoven, Committer: Simon Dickhoven; The sign-off is missing.

😊 do you want me to go back and fix those commits? i've never done that but i'm sure i can figure it out. 🙂

alpeb:
> 😊 do you want me to go back and fix those commits? i've never messed with git history but i'm sure i can figure it out.
 
Yes please, you'll have to do an interactive rebase I think. So the force-commit will be required after all.

sdickhoven:
ok. that seems to have worked. 😌 

now that i know how to edit past commits, would you like me to also prepend my commit messages with e.g.

```
fix(linkerd-cni): ...
```

?

... though that would require another force-push.

alpeb (approved):
Thanks again, LGTM!

sdickhoven:
hi @alpeb 👋 

any idea when you are going to create a new release / docker image for the cni plugin with the updated `install-cni.sh`?

sdickhoven:
> hey @sdickhoven just a heads up that @alpeb is out of the office this week, but we can likely create a new release once he is back.

thanks, @cratelyn 🙏 

no rush. ☺️ 

i just wanted to get an idea of _roughly_ when we can expect a release. that's all.

happy to wait until @alpeb is back from vacation (and longer too). 🏝️

PR #483, opened 2025-02-18
build(deps): bump dev to v45

alpeb:

PR #484, opened 2025-02-18
test(deps): use setup-rust in cni-plugin-integration/repair-controller

alpeb:
The test was failing with:

```
  cargo:warning=Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `clang-19` installed?
```

Using linkerd/dev/actions/setup-rust ensures the proper dependencies are set up.

Also, pinned to ubuntu-22.04 for now as ubuntu-latest is lacking everything this build requires.

### DIFF
--- a/.github/workflows/cni-plugin-integration.yml
+++ b/.github/workflows/cni-plugin-integration.yml
@@ -39,7 +39,7 @@ jobs:
     timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
-      - uses: linkerd/dev/actions/setup-tools@v45
+      - uses: linkerd/dev/actions/setup-rust@v45
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
       - name: Run repair-controller tests
         run: just cni-repair-controller-integration
--- a/.github/workflows/cni-plugin-integration.yml
+++ b/.github/workflows/cni-plugin-integration.yml
@@ -19,7 +19,7 @@ jobs:
         cni: [flannel, calico, cilium]
         iptables-mode: [legacy, nft]
     timeout-minutes: 15
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: linkerd/dev/actions/setup-tools@v45
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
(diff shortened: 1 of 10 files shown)


.github/workflows/cni-plugin-integration.yml
```diff
@@ -37,8 +38,9 @@ jobs:
         run: just cni-plugin-test-ordering
   repair-controller:
     timeout-minutes: 15
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
olix0r, line 41: I think we can use 24.04 if we're using v45?

alpeb: Unfortunately not. I've tried merging here my branch from #483 and using ubuntu-24.04, and it fails with the unhelpful message
```
error: could not compile `linkerd-cni-repair-controller` (bin "linkerd-cni-repair-controller") due to 1 previous error
```
https://github.com/linkerd/linkerd2-proxy-init/actions/runs/13414955803/job/37473425884?pr=484

olix0r: Ok it seems fine to me... but I'm worried that we're hiding a problem that will come back to bite us.

The underlying issue is:

```
   Compiling linkerd-cni-repair-controller v0.1.0 (/home/runner/work/linkerd2-proxy-init/linkerd2-proxy-init/cni-repair-controller)
   Compiling linkerd-network-validator v0.1.2 (/home/runner/work/linkerd2-proxy-init/linkerd2-proxy-init/validator)
error: linking with `cc` failed: exit status: 1
  |
  = note: /usr/bin/ld: /home/runner/work/linkerd2-proxy-init/linkerd2-proxy-init/target/x86_64-unknown-linux-musl/debug/deps/libopenssl_sys-445faf95f9f4260c.rlib(libdefault-lib-rand_unix.o): in function `atoi':
          /usr/include/stdlib.h:483:(.text.ossl_pool_acquire_entropy+0x15a): undefined reference to `__isoc23_strtol'
          /usr/bin/ld: /usr/include/stdlib.h:483:(.text.ossl_pool_acquire_entropy+0x181): undefined reference to `__isoc23_strtol'
          collect2: error: ld returned 1 exit status
          
  = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
```

I suspect that we're going to need to resolve this sooner than later, but I don't think we have to block CNI changes on that.

In that case, we probably want to pin _all_ of this repos's action usage to usage 22.04 to maintain consistency.

alpeb: I tried with other compiler tweaks but no luck. So I've pinned all the actions to ubuntu-22.04, and this PR should be ready to go.

zaharidichev (approved):
LGTM

PR #531, opened 2025-06-12
chore(validator): enable signal handling for Windows

zaharidichev:
This change modifies the signal handling in the network validator in order to enable it to run on Windows

Signed-off-by: Zahari Dichev <zaharidichev@gmail.com>

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -83,7 +83,6 @@ async fn main() {
         }
 
         // If the process is terminated by a signal, exit with an error.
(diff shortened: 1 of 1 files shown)


alpeb (approved):
Nice! (and funny how windows does signal handling as if it were keyboard events :shrug: )
Another nit: given the signal handlers do their own termination via `error!`, I think the `exit(UNSUCCESSFUL_EXIT_CODE)` in the `handle_signals()` branch, will never get executed.

PR #532, opened 2025-06-16
build(validator): modify Ci job to release windows binaries

zaharidichev:
This PR augments the validator release action to also publish windows binaries.


justfile-rust
```diff
@@ -52,4 +57,4 @@ package: build
 
 build *flags:
     {{ _cargo }} fetch --locked
-    {{ _cargo }} build --workspace -p {{ crate }} {{ flags }}
+    {{ _cargo }} build {{ if crate == "" { "--workspace" } else { "-p " + crate } }} {{ flags }}
```
zaharidichev, line 60: We do that to avoid compiling the whole workspace when we are just targeting one of the repos. One of the problems that happens if this is not in place is that when we are compiling for Windows systems, we try to compile the CNI repair controller as well which has some dependencies that cannot be compiled for Windows, making the whole build job fail.

PR #554, opened 2025-07-29
test(repair-controller): remove unneeded step

alpeb:
The `repair-controller` step in CI is building the cni-repair-controller separately, but that's not used directly. It is built as part of the cni-plugin image.


alpeb:
Closing in favor of #553

PR #696, opened 2026-03-18
add proxyInit dual-stack conditional

panic-kbutton:
when running Linkerd in dual stack mode today, skipSubnets breaks proxyInit with error similar to:
```
time="2026-03-18T00:21:37Z" level=info msg="/usr/sbin/ip6tables-nft -t nat -A PROXY_INIT_REDIRECT -p all -j RETURN -s <ipv4-cidr-here> -m comment --comment proxy-init/ignore-subnet-<ipv4-cidr-here>"
time="2026-03-18T00:21:37Z" level=info msg="ip6tables v1.8.11 (nf_tables): host/network `<ipv4-cidr-here>' not found\nTry `ip6tables -h' or 'ip6tables --help' for more information.\n"
```

this breaks when passing either ipv4 or ipv6 subnets to skip subnets (or both), where either the iptables-nft or ip6tables-nft command breaks with the wrong subnet. This PR adds a conditional to verify subnet parsing with the right scheme so that on each invocation only the right subnets are parsed

### DIFF
--- a/proxy-init/cmd/root.go
+++ b/proxy-init/cmd/root.go
@@ -195,9 +195,14 @@ func BuildFirewallConfiguration(options *RootOptions) (*iptables.FirewallConfigu
 	sanitizedSubnets := []string{}
 	for _, subnet := range options.SubnetsToIgnore {
 		subnet := strings.TrimSpace(subnet)
-		_, _, err := net.ParseCIDR(subnet)
+		_, cidr, err := net.ParseCIDR(subnet)
 		if err != nil {
 			return nil, fmt.Errorf("%s is not a valid CIDR address", subnet)
+		}
+
+		isIPv6Subnet := cidr.IP.To4() == nil
+		if isIPv6Subnet != options.IPv6 {
+			continue
 		}
 
 		sanitizedSubnets = append(sanitizedSubnets, subnet)
--- a/proxy-init/cmd/root_test.go
+++ b/proxy-init/cmd/root_test.go
@@ -122,4 +122,62 @@ func TestBuildFirewallConfiguration(t *testing.T) {
 			}
 		}
 	})
+
+	t.Run("It filters subnets by address family for dual-stack", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


panic-kbutton:
@alpeb any more info needed for this change? thanks!

proxy-init/cmd/root.go
```diff
@@ -195,11 +195,16 @@ func BuildFirewallConfiguration(options *RootOptions) (*iptables.FirewallConfigu
...
-		_, _, err := net.ParseCIDR(subnet)
+		_, cidr, err := net.ParseCIDR(subnet)
 		if err != nil {
 			return nil, fmt.Errorf("%s is not a valid CIDR address", subnet)
 		}
 
+		isIPv6Subnet := cidr.IP.To4() == nil
+		if isIPv6Subnet != options.IPv6 {
+			continue
+		}
```
unleashed, line 206: Maybe avoid silently dropping non-matched subnets?
```suggestion
		if isIPv6Subnet != options.IPv6 {
		    log.Debugf("skipping subnet %s: doesn't match current stack (ipv6=%v)", subnet, options.IPv6)
			continue
		}
```

proxy-init/cmd/root_test.go
```diff
@@ -122,4 +122,62 @@ func TestBuildFirewallConfiguration(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("Unexpected error: %s", err)
+		}
+		expectedIPv6 := []string{"fd00::/8", "2001:db8::/32"}
+		if !reflect.DeepEqual(config.SubnetsToIgnore, expectedIPv6) {
+			t.Fatalf("Expected IPv6 subnets %v but got %v", expectedIPv6, config.SubnetsToIgnore)
+		}
+	})
+
+	t.Run("It handles IPv4-only subnets with dual-stack enabled", func(t *testing.T) {
```
unleashed, line 164: We're missing the case of IPv6-only subnets with single-stack IPv4.

unleashed (commented):
Thanks for your submission, @panic-kbutton!

I just noticed a couple improvements we could have (see comments).

PR #742, opened 2026-05-04
chore(test): remove k3s version overrides

adleong:
We needed to override the version of k3s and lock it to `v1.27.6-k3s1` because newer versions of k3s did not work with Calico.  However, https://github.com/k3d-io/k3d/issues/1375 notes that as of `k3s v1.31` this is no longer a problem.

Therefore we remove this version overriding entirely and rely on the `latest` k3s channel from the devcontainer, which is currently `v1.35.1-k3s1`.


cratelyn (dismissed):
great find!

alpeb (approved):
LGTM! 😅

PR #760, opened 2026-06-05
CNI Plugin Integration Test

raykroeker:
The integration test script for the cni plugin was incorrectly exiting with status code '0' when calling the `wait_rollout` function.  Fixe `run.sh` to exit when the rollout status fails.

Refactor the calico setup into a setup.sh script for the integration test scenario from run-ordering.sh.

Wait for the calico-kube-controllers as well as the node daemonset in the correct namespace (calico-system) vs (kube-system).

### DIFF
--- a/cni-plugin/integration/run.sh
+++ b/cni-plugin/integration/run.sh
@@ -7,13 +7,12 @@ cd "${BASH_SOURCE[0]%/*}"
 SCENARIO=${CNI_TEST_SCENARIO:-flannel}
 IPTABLES_MODE=${IPTABLES_MODE:-legacy}
 
-# Run kubectl with the correct context.
+# TEST_CTX must be set in the environment prior to running this script
+kubectl config use-context "${TEST_CTX}"
+
+# Run kubectl
 function k() {
-  if [ -n "${TEST_CTX:-}" ]; then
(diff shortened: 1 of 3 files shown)


cni-plugin/integration/run.sh
```diff
@@ -1,19 +1,18 @@
...
+kubectl config use-context "${TEST_CTX}"
+
+# Run kubectl
 function k() {
-  if [ -n "${TEST_CTX:-}" ]; then
-    kubectl --context="$TEST_CTX" "$@"
-  else
-    kubectl "$@"
-  fi
+   kubectl "$@"
```
alpeb, line 15: Why this change? Previously a missing `TEST_CTX` was handled gracefully; with this change `kubectl config use-context` will fail.

alpeb (approved):
Looks great to me! Just one small observation below.

cni-plugin/integration/run.sh
raykroeker: Looking at all of the places it was called (w/in the justfile) TEST_CTX was always provided.  So the branching was not required.

It make the script simpler to understand that TEST_CTX was not optional, and we set it for the entire run.

PR #761, opened 2026-06-05
feat(cni-installer): implement cni installer in go

raykroeker:
This change implements improved reliability of the installer.  It removes race conditions using a single thread to handle file-system events,  leverages a shared library for those monitoring and handling those events, and provides unit test coverage.

The interface consists of a new command binary and Installer type used to perform both install and removal:

- `cni-install/main.go:` Runs the installer as well as the removal on exit.
- `pkg/cni/install.go:` Defines the factory and interface for the Installer type.

### Install

The implementation walks through 3 initial steps:

- Copy binaries to a target host path.
- Write a kube config file from a template, injecting a number of variables from the environment.
- Edit CNI configuration file(s). Injecting the linkerd cni plugin at the tail end of the config.

After the completion of these steps, the installer establishes a watch on:

- The parent directory of the api service account token. Any changes to the `..data` directory trigger a reconfiguration of the kube config file.
- The parent directory of the CNI configuration tree. Changes to the files trigger a reconfiguration of the file (in the event the file changes).

### Remove

Changes to files are tracked in the installer's in-memory log.  When the command exits/quits (via `signal.Notify`) each log event is reverted:

 - Installed binaries are removed.
 - The kube config file is removed.
 - The CNI config file changes are reverted.


cni-plugin/integration/manifests/calico/linkerd-cni.yaml
```diff
@@ -120,6 +120,7 @@ spec:
         kubernetes.io/os: linux
       hostNetwork: true
       serviceAccountName: linkerd-cni
+      automountServiceAccountToken: false
```
raykroeker, line 129: Changes to the test-scenario files drop the service account token's lifecycle down to 10m (a hard floor).

internal/cni/conf.go
```diff
@@ -0,0 +1,311 @@
...
+const (
+	// k8sConfigTemplate is a go template used to reconfigure the kubeconfig
+	// file for the linkerd plugin.
+	k8sConfigTemplate = `# kubeconfig file for linkerd.io/cni-plugin
+apiVersion: v1
+kind: Config
+clusters:
+- name: local
+  cluster:
+    server: {{ .ServiceProtocol }}://{{ .ServiceHost }}:{{ .ServicePort }}
```
raykroeker: I don't like protocol being configurable.  This is how the previous installer was implemented.  I'd like to remove this.

internal/cni/conf.go
```diff
@@ -0,0 +1,311 @@
...
+// the kubeconfig file for the linkerd plugin.
+type k8sConfigData struct {
+	// authToken is the authentication token used when invoking k8s API for the
+	// linkerd user.
+	AuthToken string
+	// CertificateAuthorityData is a base64 encoding of certificate authority
+	// data.
+	CertificateAuthorityData string
+	// SkipTLSVerify sets tls config to be insecure.
+	SkipTLSVerify bool
```
raykroeker: I don't like skip tls verify being configurable. This is how the previous installer was implemented. I'd like to remove this.

cni-install/main.go
```diff
@@ -0,0 +1,37 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+
+	ch := make(chan os.Signal, 1)
+	go func() {
+		<-ch
+		cancel()
+	}()
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
+
+	logrus.SetLevel(logrus.DebugLevel)
```
alpeb: We should probably parametrize the log level through an arg

alpeb (commented):
Thanks for this @raykroeker! I've done a first pass over the implementation and it looks pretty solid 👍

Before we move on, have you had a chance to investigate the directory-based config in CNI v1.3.0? Knowing how well that mode is supported, together with our K8s and platform support matrix, would let us decide whether to support only that mode (simpler, since it apparently avoids watching and mutating third-party config files) or to keep the mutating mode for now.

raykroeker:
> Thanks for this @raykroeker! I've done a first pass over the implementation and it looks pretty solid 👍
> 
> Before we move on, have you had a chance to investigate the directory-based config in CNI v1.3.0? Knowing how well that mode is supported, together with our K8s and platform support matrix, would let us decide whether to support only that mode (simpler, since it apparently avoids watching and mutating third-party config files) or to keep the mutating mode for now.

I haven't yet looked into that.  I will investigate write it up.

cni-install/main.go
raykroeker: Done: 96536c6ee81a806f96aa0b003783763951fc8470

alpeb:
For record-keeping purposes, let me state we're keeping on mutating the master config file, given the directory-based approach hasn't been adopted by containerd/k8s.

internal/cni/run.go
alpeb, line 1: In the [install-cni.sh](https://github.com/linkerd/linkerd2-proxy-init/blob/cni-plugin/v1.6.8/cni-plugin/deployment/scripts/install-cni.sh#L340-L346) script we are setting up the file monitoring first, and then touching the `conflist` file so it triggers an event that will inject linkerd's config. As opposed to doing an initial linkerd config injection, and then setting up file monitoring, which opens a window between those two steps where undetected changes can sneak in.

internal/cni/remove.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		if pluginType, ok := plugin[cniKeyType]; ok && pluginType == cniValTypeLinkerd {
+			linkerdAt = i
+			break
+		}
+	}
+	if linkerdAt >= 0 {
+		plugins = append(plugins[:linkerdAt], plugins[linkerdAt+1:]...)
+		val[cniKeyPlugins] = plugins
+		data, err = json.Marshal(val)
```
alpeb: Can you use `json.MarshalIndent` here instead? I noticed while testing shutdown that we end up with a one-line json file while the original one (e.g. deployed by Calico) was indented.

internal/cni/watch_fs.go
```diff
@@ -0,0 +1,96 @@
...
+
+// applies returns true if the operation is included in the set the watch is
+// looking for.
+func (fsw watch) applies(op fsnotify.Op) bool {
+	for _, wo := range fsw.operations {
+		if op.Has(wo) {
+			return true
+		}
+	}
+	return false
```
alpeb:
```suggestion
	return slices.ContainsFunc(fsw.operations, op.Has)
```

internal/cni/run.go
```diff
@@ -0,0 +1,112 @@
...
+	log.WithField("installed-files", installed).Debug("installed binary files")
+	watchOperations := []fsnotify.Op{
+		fsnotify.Create,
+		fsnotify.Rename,
+		fsnotify.Write,
+	}
+	watches := []watch{
+		{
+			eventFN: func(event fsnotify.Event) error {
+				if strings.HasSuffix(event.Name, "..data") {
```
alpeb, line 58: Please replicate here the comments from `install-cni.sh` explaining why we're watching over `..data` to detect a SA token rotation.

internal/cni/run.go
```diff
@@ -0,0 +1,112 @@
...
+				return nil
+			},
+			operations: watchOperations,
+			path:       hostCNIConfig(),
+		},
+	}
+	errs := make(chan error, 1)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	err = i.watchFS(ctx, errs, watches)
```
alpeb, line 89: In `install-cni.sh` there's a wait checking that `inotifywait` is ready to process events. Can you check if go's fsnotify implementation accounts for that when calling `Add`?

internal/cni/run_test.go
```diff
@@ -0,0 +1,189 @@
...
+				//	remove the link ..data -> auth-root
+				//	create the link ..data -> new-auth-root
+				//
+				// this will fire a create event for the auth token file's root
+				newAuthRoot := path.Join(self.root, "..new-auth-root")
+				mustMkdir(t, newAuthRoot)
+				mustCopyFile(t, newAuthRoot, "testdata/auth-token-new")
+				mustRemove(t, path.Join(self.root, "..data"))
+				mustLink(t, newAuthRoot, path.Join(self.root, "..data"))
+				mustRemove(t, path.Join(self.root, "..auth-root"))
```
alpeb, line 71: 🤌

internal/cni/installer_test.go
```diff
@@ -0,0 +1,26 @@
...
+	expSources := []string{
+		"env:CNI_NETWORK_CONFIG",
+		"file:/tmp/linkerd.yaml",
+	}
+	switch mgr := mgr.(type) {
+	case *installer:
+		if len(mgr.sources) != len(expSources) {
+			t.Fatalf("default sources list is incorrect %d<>%d", len(expSources), len(mgr.sources))
+		}
+		for i := 0; i < len(expSources); i++ {
```
alpeb:
```suggestion
		for i := range expSources {
```

cni-plugin/integration/manifests/calico/linkerd-cni.yaml
alpeb: Can you expand a bit on how this affects the tests?

internal/cni/conf.go
alpeb: sgtm, this probably has never been changed from its default.

internal/cni/conf.go
alpeb: :+1:

internal/cni/run.go
raykroeker: Creating the [watcher](https://github.com/fsnotify/fsnotify/blob/v1.10.1/backend_inotify.go#L147) creates and initializes the inotify backend.  It makes the init syscall which returns a file descriptor from which it reads events and sends events and errors back on the existing channels. Once the watcher is created, it's in the ready state.

Any issues with the read loop will close the channels and exit the cni run loop with an error.

alpeb: Awesome :+1: , thanks for the extra research.

alpeb (approved):
Looks awesome 👍 
Looking forward to getting this integrated. I think we should create a new branch off of the `cni-plugin/v1.6.8` tag, e.g. `cni-plugin/v1` for the maintenance releases consumed by linkerd 2.20, and continue the development of these new CNI plugin features in `main`, which we can start releasing as `cni-plugin/v2.0.0+alpha.1`. LMKWYT.

raykroeker:
> Looks awesome 👍 Looking forward to getting this integrated. I think we should create a new branch off of the `cni-plugin/v1.6.8` tag, e.g. `cni-plugin/v1` for the maintenance releases consumed by linkerd 2.20, and continue the development of these new CNI plugin features in `main`, which we can start releasing as `cni-plugin/v2.0.0+alpha.1`. LMKWYT.

This strategy makes sense to me, I'll set it up.

PR #762, opened 2026-06-08
fix: clean up malformed CNI error output

immanuwell:
This fixes a small but real logging bug in `cni-plugin`.

Some CNI error paths use `%e` for `error`, so Go prints `%!e(...)` junk instead of the actual message. That shows up on normal failures like bad CNI JSON, bad kube config, or missing integration mounts. Kinda noisy, and it hides the real problem.

What changed
- switch those log and test format strings to `%v` or `%s`
- add a small regression test for the invalid JSON path
- fix the `proxy-gid` mismatch message in the integration test helper

Repro
1. On `main`, run `go test ./cni-plugin/integration/tests/flannel -run TestLinkerdIsLastCNIPlugin -v`
2. In a plain local env, the expected mount is missing, so the test fails
3. Before this patch the failure contains `%!e(...)`
4. With this patch it prints the real path error

Checks
- `go test ./cni-plugin ./proxy-init/... ./pkg/...`

### DIFF
--- a/cni-plugin/main.go
+++ b/cni-plugin/main.go
@@ -144,7 +144,7 @@ func parseConfig(stdin []byte) (*PluginConf, error) {
 func cmdAdd(args *skel.CmdArgs) error {
 	conf, err := parseConfig(args.StdinData)
 	if err != nil {
-		logrus.Errorf("error parsing config: %e", err)
+		logrus.WithError(err).Error("error parsing config")
 		return err
 	}
 	configureLoggingLevel(conf.LogLevel)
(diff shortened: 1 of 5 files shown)


cni-plugin/main.go
```diff
@@ -144,7 +144,7 @@ func parseConfig(stdin []byte) (*PluginConf, error) {
 func cmdAdd(args *skel.CmdArgs) error {
 	conf, err := parseConfig(args.StdinData)
 	if err != nil {
-		logrus.Errorf("error parsing config: %e", err)
+		logrus.Errorf("error parsing config: %v", err)
```
raykroeker: Instead of logging errors with a format string (`Errorf`) please use `WithError(err).Error(...` instead.  Here and below as well.

raykroeker (changes requested):
Thanks @immanuwell for the PR.  Instead of using format strings, please use the structured error field on the logger.

cni-plugin/main.go
immanuwell: done

immanuwell:
@raykroeker done, addressed all the requested changes, PTAL

alpeb (commented):
Thanks @immanuwell , the main change looks good. I don't think you should add a unit test for logging though; it tests the logging framework not our behavior and I can see this sort of thing can become a burden to maintain when the framework changes.

PR #775, opened 2026-06-30
fix(test): make cni-repair-controller test more resilient

alpeb:
The test was failing when deploying the pause DS and immediately checking for its readiness, as the kube-api didn't have time to acknowledge the existence of that workload. Added a loop around that, as well as homogenizing the last for loop checking for DS replica replacement.


raykroeker (commented):
If the pause-app never meets the `ready` condition, what if anything times out the wait loops?

alpeb:
> If the pause-app never meets the `ready` condition, what if anything times out the wait loops?

The repair-controller job has a timeout of 15m

raykroeker (approved):

PR #784, opened 2026-07-07
feat(cni-installer): integrate cni installer

raykroeker:
Build cni installer into the runtime docker image.

Drop tooling from the runtime docker image required to enable the installer shell script. Remove the shell script and jq filter file from the repo.

Tested by running through the plugin integration tests locally.


alpeb (approved):
LGTM! 👍

